PR #3, opened 2018-09-21
Fix double backslash issue

cleemullins:


cleemullins:
Reviewed in person dith Darren

PR #87, opened 2018-10-10
add stylecop rules and packages

lzc850612:
# Problem

Need to add stylecop to our build

# Solution

Added AsyncUsageAnalyzers and StyleCop.Analyzers components to all projects
Added AsyncUsageAnalyzers and Microsoft.CodeAnalysis.CSharp.Features (StyleCop.Analyzers) rules


lzc850612:
Build validation failed because of an issue in the VSTS build definition. That's fixed so we can ignore this failure.

lauren-mills (approved):

Issue #118, opened 2018-10-16
Add launchSettings.json to .gitignore


gabog:
In a team development launchSettings.json is where devs normally override configuration settings to be able to run the solution locally without having to modify appsettings.json. 

It would be nice to add this file to gitignore to avoid accidently committing changes with local settings and to help devs keep their own configuration local.

This will also help contributors that work on their own branches with their own settings (I normally override the path to the bot file and my local settings in launchSettings.json.

The following line in gitignore should achieve this:
**/Properties/launchSettings.json

darrenj:
We reviewed this change - if we don't have the launchSettings file under source control, it would be generated with a random port which is different from the deployment script which will make developer's local run fail out of the box which is important. We should still keep this file and just be sure to not include it in any changes. You can always add changes on top of it to override any settings locally.****

gabog:
ok

PR #124, opened 2018-10-16
add launchsettings to git ignore and update the default launchsetting

lzc850612:
# Problem

launchSettings.json should be something developer can override and not worry about accidentally commit. 

# Solution

Add launchSettings.json file to gitignore
Fix default launchSettings for Virtual Assistant to be the same as the bot file.


lzc850612:
#118

lzc850612:
Lauren mentioned that if we don't have the launchSettings file under source control, it would be generated with a random port which is different from the deployment script which will make developer's local run fail. We should still keep this file and just be sure to not include it in any changes. You can always add changes on top of it to override any settings locally.

darrenj (approved):

Issue #129, opened 2018-10-17
Consolidate AzureMapsKey/AzureMaps-Key usage


ryanisgrig:
**Description**
Standardize `AzureMapsKey` usage in the POI skill. Documentation refers to it as `AzureMaps-Key` but code uses `AzureMapsKey`. We will stick to `AzureMapsKey` everywhere.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

Issue #121, opened 2018-10-16
PoI skill should detect and warn with a specific exception that the AzureMaps key isn't provided (see main Bot startup)


darrenj:
**Description**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

ryanlengel:
![image](https://user-images.githubusercontent.com/43043272/47112048-0cbcb600-d20a-11e8-9d01-29ee6159090b.png)
The skill catches it when you try to invoke it & returns with a trace error. Should it warn before invoking (on startup)?

PR #132, opened 2018-10-17
AzureMapsKey checking & consolidation

ryanisgrig:
#121 Throw specific exception for missing AzureMapsKey string
#129 Removed all AzureMaps-Key references


solutions/Virtual-Assistant/src/csharp/skills/pointofinterestskill/Dialogs/Shared/PointOfInterestSkillDialog.cs
```diff
@@ -370,7 +370,7 @@ protected string GetAzureMapsKey()
         {
             _services.Properties.TryGetValue("AzureMapsKey", out var key);
 
-            if (key == null)
+            if (string.IsNullOrEmpty((string)key))
```
lzc850612: Better to define a variable for key so you can reuse it in the else branch:

var keyStr = (string)key;
if (string.IsNullOrWhiteSpace(keyStr))
{
    ...
}
else
{
    return keyStr;
}

lzc850612 (approved):

PR #138, opened 2018-10-18
fix some bugs

feich-ms:
Fix below issues:
1. Prompt a message("Your to do list is empty.") to users if users try to delete/mark a task of an empty task list. Before fixing, it always prompted "Which task will you operate" which is not good experience.
2. Add some new patterns for "AddToDo", including "remind me to {TaskContent}" and "remind me that {TaskContent}".
3, Add a new entity type called "TaskType" which will be used later to distinguish task type(ToDo, Shopping, Grocery).

### DIFF
--- a/solutions/Virtual-Assistant/src/csharp/skills/todoskill/Dialogs/Shared/Resources/ToDoSharedResponses.de.json
+++ b/solutions/Virtual-Assistant/src/csharp/skills/todoskill/Dialogs/Shared/Resources/ToDoSharedResponses.de.json
@@ -282,5 +282,14 @@
       }
     ],
     "inputHint": "acceptingInput"
+  },
+  "NoTasksInList": {
+    "replies": [
+      {
+        "text": "Ihre To Do Liste ist leer.",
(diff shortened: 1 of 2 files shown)


darrenj (changes requested):
Hi,

Changes look good but the responses should be added to the .json file in the dialog specific directory (e.g. Dialogs\DeleteToDo\Resources) rather than shared unless truly shared. Also we need to ensure the language localised versions are kept up to date otherwise we'll get bugs when demoing in Chinese for example.

For all new or modified text responses can we make sure we update the EN, DE and ZH JSON response files (where they are present - I know there are some gaps). Use Bing Translate for German to provide an initial version for anything new you add.

@lauren-mills fyi

feich-ms:
Thanks to @darrenj for your nice suggestions. For your questions:
1. The added response is shared by DeleteToDo and MarkToDoCompleted flows, so I put it into the shared response folder.
2. I really forgot to add corresponding response in DE and ZH json file. I have sent a new commit and added the response in DE and ZH files. Thanks for the reminder.

darrenj (approved):

Issue #165, opened 2018-10-22
Do not throw exception if TranslatorKey for TranslationMiddleware is missing


ryanisgrig:
**Description**
If building a Virtual Assistant off of the master branch - the startup code is expecting a Translator API key that is not included in deployment.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

PR #167, opened 2018-10-22
Rylengel/disabletranslatorstartup

ryanisgrig:
If translatorKey is missing - will continue ahead rather than break with an exception. This leave it optional for the developer to provide if they so choose.


lauren-mills (approved):

lzc850612 (approved):

ryanisgrig:
#165

PR #173, opened 2018-10-24
Updated T4 templates 

gabog:
Hi, I updated the T4 templates with a new version that dynamically determines the location of the json file during code gen so we don't need to alter the template for each new resources we create.

Let me know if you have any issues or questions


darrenj:
@lzc850612 made some changes yesterday which are in master around adding better exception handling to handle files JSON files with errors - you'll likely have to do a merge?

https://github.com/Microsoft/AI/pull/168/files

darrenj (changes requested):
@lzc850612 made some changes yesterday which are in master around adding better exception handling to handle files JSON files with errors - you'll likely have to do a merge?
https://github.com/Microsoft/AI/pull/168/files

lzc850612 (commented):
Gabo, T4 template has a way of include a file: https://docs.microsoft.com/en-us/visualstudio/modeling/t4-include-directive?view=vs-2017.

With this feature we can define the template all in one file, ResourceTemplate.tp and put it in Microsoft.bot.solutions project, then in the *.tt file, we only need to do:

<#@ template debug="false" hostspecific="true" language="C#" #>
<#@ output extension=".cs" #>
<#
  var resourcePath = System.IO.Path.GetDirectoryName(Host.TemplateFile);
  var projectPath = Host.ResolveAssemblyReference("$(ProjectDir)");
  var relativeResourcePath = resourcePath.Substring(projectPath.Length);
#>

<#@ include file="..\..\..\..\..\microsoft.bot.solutions\Resources\ResponseTemplate.tp" #>

and in the .tp file, just use relativeResourcePath. this way we don't have to have all the duplicated .tt file in all the places and it'll be so much easier to make changes in the future.

Feel free to apply this in your current change, or if you want to merge your change first, i can make that change after yours is merged. My previous change was only on a few .tt files (which i think you've merged) which is part of the reason that got me thinking about this file inclusion approach to apply something on all T4 files together. (just thought there's gotta be a better way!)

gabog:
Hi Ted, I didn't know about the include directive, let me play with it a bit and I will update the PR if I can figure out the best place to put the shared file.

lzc850612:
@gabog sure thing. I think we can put it in microsoft.bot.solutions project, under Resources. Then it'll be able to be referenced by all skills.

gabog:
Hi @lzc850612 and @darrenj . I updated the code to use a shared t4 template approach you suggested.

In addition to that, I also:
	• Extracted the logic to read and create responses out of the generated code an put it in a separate ResponseManager class. 
	This allowed me to create unit tests and address some of the common issues we have more explicitly.
	It will also allow devs to use a different resource manager in the future if they want without having to regenerate the code.
	• I also change the logic so we don't assume English as the fallback language. So I renamed all the .en.json files we have as .json, these will be the default and it doesn't need to be english, other people can assume the default is spanish and add en.json later. (Resx style).

Please review and let me know what you think.

Gabo

gabog:
btw, I added 8 more tests but they don't show in the build results, if you approve the PR, you may need to update the VA-PRValidation to include the new tests for Microsoft.Bot.Solutions.Tests.

Gabo

lzc850612:
> btw, I added 8 more tests but they don't show in the build results, if you approve the PR, you may need to update the VA-PRValidation to include the new tests for Microsoft.Bot.Solutions.Tests.
> 
> Gabo

Thanks Gabo. I think it's probably because of the test filter i put in the build definition. I've just made the change so it'll run the tests after your change is merged. Or if you need to push a new iteration, it'll trigger another build so it'll pick up my change the run your tests.

solutions/Virtual-Assistant/src/csharp/skills/calendarskill/Dialogs/Shared/Resources/CalendarSharedResponses.cs
```diff
@@ -1,124 +1,44 @@
...
-                        _jsonResponses = null;
-                        resourceLoadingException = new Exception($"Deserialization exception when deserializing response resource file {file}: {ex.Message}");
-                        break;
-                    }
-                    catch (Exception ex)
-                    {
-                        _jsonResponses = null;
-                        resourceLoadingException = ex;
-                        break;
-                    }
```
lzc850612, line 74: These try catch are added by me for the shared responses in four skills. I didn't put it in all response.cs files but i should. Can you add this into the common .t4 file? it's needed for better error handling.

lzc850612: actually let me add them. i'll have a better idea how to test it.

lzc850612 (approved):

gabog:
OK. Thanks. I saw those but couldn’t come up with a test that required that try/catch and I didn't port it after refactoring everything to Response Manager.

Sent from mobile
________________________________

darrenj (approved):

PR #176, opened 2018-10-24
added customer support template

### LINKED ISSUE #189: Retail Customer Support luis model

lauren-mills:
fixes #189, #188, #187, #186, #185, #184, #183


darrenj (changes requested):
Hi Lauren,

A few observations that would be good to try and incorporate, nothing major.

- [x] WebChat doesn't sent the conversationUpdate message so when people test (internally to start) they don't get the introduction card which makes it hard to get started. Can you ask William for a code example that sends the conversationUpdate event as part of the onLoad. This should be something WebChat looks to add by default really

- [x] Can we add suggested actions to the Intro Card and Help Card with example questions otherwise it's hard for (testers) to get started.

- [x] Remove ConfirmYes, ConfirmNo from General.Lu (and we should go back to EnterpriseTemplate too

- [x] Retail.lu not in the CognitiveModels folder (LUIS model in the deploymentScripts)

- [x] What is your ZipCode prompt on Exchange item doesn't say "Sorry. that doesn't look like a valid zip code" but uses the same prompt name

- [x] I found it hard to trigger the CheckItemAvailablity perhaps consider adding a pattern along the lines of "Do you {item} in stock" or {I'd like to buy {item} do you have it in stock"?

- [x] I typed can I buy and collect in store and it asked if I wanted to put the item on hold. I didn't give one ;-) 

- [x] The exception caused by clicking in the update info adaptive card causes the bot to be "stuck" in the dialog. Tweak exception handler? Cancel rectified it

- [x] "update my delivery address" or "change my shipping address" doesn't trigger the right intent

- [ ] One for the next revision likely but we should add localised responses for the dialog responses (use bing translator). Raise an issue and we can track

darrenj (approved):

PR #194, opened 2018-10-26
DO NOT MERGE fix confirm intent bug

Theodosia1005:


darrenj (changes requested):
Yes, No and More are no longer intents we support in the General Model. Use ConfirmPrompts and the Next intent.

lauren-mills:
Closing. The general model will no longer have the ConfirmNo, ConfirmYes, and More intents and will instead use the ConfirmPrompt, and the Next intent to cover these scenarios.

PR #200, opened 2018-10-29
poi fix for incorrect locale

lauren-mills:
Found that the default locale of "en" in POI resulted in API errors. These were fixed when switching to the full 4 character code "en-us"


lzc850612 (approved):

ryanisgrig (commented):
nit: should this string be moved to the resources file?

lauren-mills:
I'm not sure.... I'm thinking it might be confusing to combine text responses with config values, so maybe appsettings? We could send it across from the VA in the configuration values with the Azure Maps key

lzc850612:
I think what Ryan is saying is to put the constant strings to a separate static class. we shouldn't put them in the response file because they're not something that needs localization. But we still will benefit if we define it in one place and use it from other.

ryanisgrig:
Yes, Ted is correct. This particular locale is a default locale to Azure Maps and isn't something users would be expected to configure.

lauren-mills:
I think we can find a more elegant solution for this in the future as we continue refining localization. Going to merge the bug fix as is right now and revisit later

PR #214, opened 2018-11-02
Update calendar luis model

KayMKM:
add some utterances and fix some wrong labels.

The test cases have been passed:

- show my meetings today
- create a meeting with jing li at 10pm for half an hour
- schedule a meeting on Nov. 3 from 3-4pm
- Update my calendar from 7 pm to 6 pm today
- move my 10pm meeting to 9pm
- cancel my appointment at 9pm

These languages' .lu files had been updated:

- de
- en
- es
- fr
- it


darrenj (changes requested):
LUIS model changes need to be made to the corresponding LU file for the Skill. Alternatively you can convert a LUIS JSON file to a LU file using the ludown refresh but be aware that we now have localised LU files (including ZH) so need to ensure any chance is replicated across these otherwise we'll end up with model drift. Any LUIS model in the deploymentScripts folder should be created through a call to ludown parse ToLuis --in LUFILENAME 

Also the Dispatch model will likely need re-creating depending on the change, without this the VA won't be able to dispatch questions correctly if the models are out of sync

lauren-mills:
@KayMKM can you change this PR to not change calendar.luis file and instead update the calendar.lu file with the changes you want? I can then make sure everything else is consistent with those changes (dispatch, localization, etc)

KayMKM:
@darrenj @lauren-mills the calendar.luis change has been reverted. And the new calendar.lu files are generated. But when I translate to zh-hans, an undefined error happened. It seems you met this error before. Could you help to generate it or tell me how to fix the error?

darrenj:
Thanks, you need the new version of ludown but we’ve just made changes to the zh-hans LU files. Can you skip the ZH-HANS file creation for now and raise an issue assigned to me to merge this in this week.

Thanks.

darrenj (approved):

darrenj:
@lauren-mills  - Can you regenerate the dispatcher with the latest changes and do a quick test pass before merging? Thanks :-D

lauren-mills:
Working on it now.

lauren-mills (changes requested):
Blocking merge until testing is done

lauren-mills:
I updated the files for dispatch.luis and calendar.luis in the msbotClone folders for en and zh deployments

lauren-mills (approved):

PR #215, opened 2018-11-02
fix update and delete time bug

Theodosia1005:
## Description

Fix find event by time's Timezone not UTC bug in update and delete flow.

## Testing Steps

tested with Google Calendar.
Tested update and delete the event with start time.


KayMKM:
cherry picked to #216

PR #217, opened 2018-11-02
Feich/bug fix

feich-ms:
Optimization
## Description
1. Remove "OneNote" in response file.
2. Remove IOutlookService.cs file which is not used any more.
3. Update the content in appsetting.config to align with code


youyu16 (approved):

feich-ms:
Commits before 5223124 are from last PR, not real changes. This PR only changed little code.

PR #227, opened 2018-11-02
Add Cortana integration features for enterprise template

AlfredoMS:
The general objective is to make the enterprise template Coranta ready.
For this purpose I have added speech support, inputHints and extra code to handle Cortana's launch intent.

## Description
- Add support for Cortana launch (when invioked via "ask [name of my skill]"
- Add suppor for speach/ssml to all messages including text, adaptive cards and hero cards. Also those one generated by TemplateManager
- Readjusted flow such as it makes more sense when running in cortana while not really affecting other channels. This means the Onboarding scenario will run before showing the intro card the first time the skill is used.

## Related Issue
internal VSTS bug if needed

## Testing Steps
Created a skill using this template's code.
Register a Cortana channel for the bot
Invoke via Cortana in Windows


AlfredoMS:
@lauren-mills

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Onboarding/OnboardingDialog.cs
```diff
@@ -52,7 +53,7 @@ public async Task<DialogTurnResult> AskForName(WaterfallStepContext sc, Cancella
             {
                 return await sc.PromptAsync(NamePrompt, new PromptOptions()
                 {
-                    Prompt = await _responder.RenderTemplate(sc.Context, "en", OnboardingResponses._namePrompt),
+                    Prompt = await _responder.RenderTemplate(sc.Context, "en", OnboardingResponses._namePrompt, inputHint: InputHints.ExpectingInput),
```
lauren-mills: For prompts, the input hint is set to expecting automatically by the BotBuilder SDK [here](https://github.com/Microsoft/botbuilder-dotnet/blob/16b211eec8749f2d576832d389556745399f8fdc/libraries/Microsoft.Bot.Builder.Dialogs/Prompts/Prompt.cs#L47), so I don't think this piece is necessary

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Escalate/Resources/EscalateStrings.resx
```diff
@@ -118,6 +118,6 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="PHONE_INFO" xml:space="preserve">
-    <value>Our agents are available 24/7 at 1(800)555-1234.</value>
+    <value>Our agents are available 24/7 at 1(800)555-1234. Or connect with us through Microsoft Teams.</value>
```
lauren-mills, line 121: Is this addition strictly necessary? Not all users of the Enterprise Bot template will necessarily have Teams integration

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Main/MainDialog.cs
```diff
@@ -34,14 +35,16 @@ protected override async Task OnStartAsync(DialogContext innerDc, CancellationTo
...
 
-        var view = new MainResponses();
-        await view.ReplyWith(innerDc.Context, MainResponses.Intro);
-
         if (string.IsNullOrEmpty(onboardingState.Name))
         {
             // This is the first time the user is interacting with the bot, so gather onboarding information.
             await innerDc.BeginDialogAsync(nameof(OnboardingDialog));
         }
+        else
```
lauren-mills: I think this should be changed back to the original flow, with ignoringInput set on the intro card

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Main/MainDialog.cs
```diff
@@ -132,7 +135,10 @@ protected override async Task RouteAsync(DialogContext dc, CancellationToken can
     protected override async Task CompleteAsync(DialogContext innerDc, CancellationToken cancellationToken = default(CancellationToken))
     {
         // The active dialog's stack ended with a complete status
-        await _responder.ReplyWith(innerDc.Context, MainResponses.Completed);
+            await _responder.ReplyWith(innerDc.Context, MainResponses.Completed, inputHint: InputHints.IgnoringInput);
+
+            var view = new MainResponses();
```
lauren-mills: Why are you showing the intro card in the Complete method? This means the intro will show every time a dialog completes.

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Shared/RouterDialog.cs
```diff
@@ -22,6 +23,12 @@ protected override async Task<DialogTurnResult> OnContinueDialogAsync(DialogCont
         {
             var activity = innerDc.Context.Activity;
 
+            if (CortanaHelper.IsLaunchActivity(activity) && string.IsNullOrEmpty(activity.Text))
```
lauren-mills, line 29: There might be a way to include a helper method to check if the message is the first message independent of channel, so that we don't have to have a channel specific check here. We can still merge this in for now

lauren-mills (changes requested):

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Escalate/Resources/EscalateStrings.resx
AlfredoMS: As an enterprise demos we want to display what is possible, this doesn't mean developers will use everything. Adding a button for MS-Teams was a requirement I had so I think it makes sense to update the text as well

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Main/MainDialog.cs
AlfredoMS: the reason i need to change this is because it makes more sense in a voiced enabled context.
If we show the main card first, it will immediatly disappear because we do the first time onboarding. Instead, the first time this bot is run we do the onboarding and at the end we show the main card.

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Main/MainDialog.cs
AlfredoMS: I wanted this to still be displayed After doing the onboarding.

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Onboarding/OnboardingDialog.cs
AlfredoMS: ok, that is good to know. I will remove them

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Shared/RouterDialog.cs
AlfredoMS: I agree this is not great. Ideally, Cortana should send and Activity of type "launch", however it currrenlty doesn't work like that so it has to be channel specific

AlfredoMS:
Pending from PR to remove onboarding from Main dialog and move it into an action within the main card

lauren-mills (approved):

PR #237, opened 2018-11-06
Changed customassistant references to virtualassistant in URLs

paragppp:
## Description
All URLs in Intro jsons seem to be referring to customassistant in the URL which seems to have changed to virtualassistant. Changed them to virtualassistant.

## Related Issue
#238 

## Testing Steps
After starting the first conversation, when clicking on "Linked Accounts" or "Skills" we get 404


darrenj (approved):
Good catch, Thanks. (The joy of name changes!)

Issue #199, opened 2018-10-29
Remove default lat-lon current coordinates from POI state


ryanisgrig:
**Description**
Remove the default lat-lon pair of CurrentCoordinates and gracefully handle missing current coordinates when POI skill is invoked.

**Expected behavior**
VA will require a location event sent to work properly and raise an exception if missing, similar to missing maps api key

PR #250, opened 2018-11-07
Remove default coordiantes for POI state

ryanisgrig:
## Description

I've removed the default coordinates in POI, a user will now be required to pass on their own `IPA.Location` event.

## Related Issue

#199 

## Testing Steps

**New Conversation**
User: find me directions to Starbucks
Bot: *throws exception, trace shows current coordinates are missing&

**New Conversation**
User: /event:{ "Name": "IPA.Location", "Value": "34.05222222222222,-118.24277777777778" }
User: find me directions to Starbucks
Bot: here's what I found for you...

### DIFF
--- a/solutions/Virtual-Assistant/src/csharp/skills/pointofinterestskill/PointOfInterestSkillState.cs
+++ b/solutions/Virtual-Assistant/src/csharp/skills/pointofinterestskill/PointOfInterestSkillState.cs
@@ -17,7 +17,7 @@ namespace PointOfInterestSkill
             DialogName = string.Empty;
             SearchText = string.Empty;
             SearchAddress = string.Empty;
-            CurrentCoordinates = new LatLng() { Latitude = 47.640568390488625, Longitude = -122.1293731033802 };
+            CurrentCoordinates = null;
             FoundLocations = null;
             ActiveRoute = null;
             SearchDescriptor = string.Empty;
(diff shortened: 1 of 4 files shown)


solutions/Virtual-Assistant/src/csharp/skills/pointofinterestskill/PointOfInterestSkillState.cs
```diff
@@ -63,5 +63,13 @@ public void Clear()
             FoundLocations = null;
             LastUtteredNumber = null;
         }
+
+        public void CheckForValidCurrentCoordinates()
+        {
+            if (CurrentCoordinates == null)
+            {
+                throw new Exception("The bot state is missing any current coordinates. Make sure your event architecture is correctly configured.");
```
lzc850612, line 71: This exception message is going to be exposed to the user right? I feel like we should show something like 'The bot doesn't have current location setup and please contact bot developer to fix it' or something like it. 

In longer term we should probably have different types of Exceptions defined and render them in a more user friendly way that the user can understand and action upon and log detail exceptions for developer to see.

ryanisgrig: This will be exposed to the developer in the emulator, the end-user will get a generic "something went wrong" exception (this is the same as if the Azure Maps key is missing).

I agree that we should have different types of exceptions, it could be useful for me as an end user to know that specifically the bot has never received GPS coordinates from my vehicle.

lzc850612 (approved):

darrenj:
Agree with removing this. We might want to consider a message back to the user highlighting that the skill hasn't  been passed any location information as part of the skill invocation. That way in demos we don't get a "sorry, I can't do that" type of message. The emulator isn't always used and this could lead to confusion.

PR #277, opened 2018-11-12
Provide the ability to control userId in LinkedAccounts

darrenj:
## Description
Converted the UserId display to a TextBox enabling the user to paste in a different userId supporting demos in the Emulator (fixed UserId) and the new webchat ahead of it "logging in" and using the right userid.

## Testing Steps

- Start Linked Accounts
- Change the UserId (to anything)
- click the button, signin (should go green)
- change the userid to something else - signin circle should not be green
- switch back to previous userid - now green


ryanisgrig (approved):
Good idea!

PR #282, opened 2018-11-12
Updated inputHint to ignoringInput

gabog:
## Description

- Updated inputHint to ignoringInput so the mic doesn't get open before the next prompt. 
- Updated solutions project so CommonResponses.tt doesn't get copied to the output folder.

## Testing Steps
using the microphone say:

1. Find a starbucks
2. Get option 1
3. The bot should not open the mic until the Shall we get started prompt is rendered.

For the tt file, simply rebuild the solution and ensure that there are not .tt files in the bin folders.


gabog:
@ryanlengel, FYI, there are several responses with expectingInput where I think we may need to use ignoringInput but I didn't change them because I don't know all the execution paths, please check and let me know if we need to change more.

darrenj (approved):

PR #290, opened 2018-11-13
updated enteprise bot to use latest telemetry + other small formattin…

lauren-mills:
## Description

Updated Telemetry and fixed other small formatting fixes


darrenj:
While we are making this changes is it a good time to bring across the base bot response localisation from VA and the refined cancellation logic (gated on a higher confidence value?)

lauren-mills:
@darrenj I updated the responses, intro cards, and and lu files to the latest. I also added the .bat files from VA

darrenj (approved):

darrenj:
@lauren-mills  - We'll need to update the Ent Template docs around deployment before we push this change

PR #296, opened 2018-11-13
Change LU design

Theodosia1005:
## Description

change entity definition in lu to fix startTime and originalStartTime confuse problem. 

1. Create a meeting from DATETIME1 to DATETIME2
2. Update a meeting from DAETTIME1 to DATETIME2

The DATETIME1 in the first sentence is labelled as StartTime, but OriginalStartTime in the second one.
The  DATETIME2 in the fist sentence is labelled as EndTime, but StartTime in the second one.
So for the same DATETIME in same position in different sentences, it can be labelled differently.
At the same time DATRETIME in different positions can be labelled with same entity.

So now for second sentence, we will get DATETIME1 being leballed as StartTime and OriginalStartTime at the same time, the flow will not change the meeting.

We change those all to from time and to time, then detect their actual usage in  DigestCalendarLuisResult part with the help of intent value.  

## Testing Steps

Change my meeting from 2pm-3pm to 4pm-5pm
change my meeting at 2pm to 4pm
delete my meeting from 2pm to 3pm
 Schedule a meeting on Nov. 3 from 3-4pm
set an denist appointment for tomorrow


solutions/Virtual-Assistant/src/csharp/skills/calendarskill/Dialogs/Shared/CalendarSkillDialog.cs
```diff
@@ -388,230 +390,319 @@ protected async Task DigestCalendarLuisResult(DialogContext dc, Calendar luisRes
...
+                        {
+                            break;
                         }
-                    }
                 }
+            }
+            catch
+            {
+                // put log here
+            }
```
lzc850612, line 632: probably shouldn't swallow the exception here?

lzc850612 (changes requested):

solutions/Virtual-Assistant/src/csharp/skills/calendarskill/Dialogs/Shared/CalendarSkillDialog.cs
KayMKM: @lzc850612 had fixed in the new commit

lzc850612 (approved):

PR #311, opened 2018-11-16
Move onboarding flow to a button in enterprise template

AlfredoMS:
## Description
Move onboarding flow to a button so it runs better on Cortana

It was agreed this flow would work for all channels


AlfredoMS:
@lauren-mills

lauren-mills (approved):

darrenj (approved):

PR #326, opened 2018-11-20
add read more intent

Theodosia1005:
## Description

Add read more to general LU


lzc850612:
is there a code change coming soon to use this language model change?

I think it's better to incorporate the models from the pre-built model team to avoid manual update on the language model because it's error prone and we currently don't have a good way to gate changes on that.

darrenj:
Seems to be an overlap with next intent - can we instead use that base intent that already exists?

bobokids:
Based on new design of the flow, when user ask to show my emails, we will show 6 emails but read the top 3. so we need two intents, ShowMore to show next 6 emails, and ReadMore to read out next 3 events.
This change is to adding the intent to the utility model, we still need to add code in skills to enable the read more feature. we need this change in so we can work on features. We will also add this intent to pre-build domain at the same time.

darrenj:
My concern is that the two intents have pretty much the same training data which will cause confusion, looking at the utterances we can just re-use the one intent?

```
# Next
- advance
- go forward
- go to next
- next
- next item
- show next
- could you show the next one
- i need to go to the next one
- go to the next one please
- move to the next one
- go on
- go to next
- go to the next page
- move forward
- move on
- move on next
- next one
- next please
- show me next
- show me the next
- show the next
- show the next one
- the next email
```
```
# ReadMore
- advance
- could you show the next one
- go forward
- go on
- go to next
- go to the next one please
- go to the next page
- i need to go to the next one
- move forward
- move on
- move on next
- move to the next one
- next
- next item
- next one
- next please
- show me next
- show me the next
- show next
- show the next
- show the next one
- the next email
```

darrenj (changes requested):
Leverage the existing ShowNext intent which has the same utterances as the ReadMore intent otherwise you'll have overlap/confusion in the LUIS model.  

Alternatives are to have very distinct utterances for a ReadMore intent but need to be careful to ensure there isn't overlap. Or alternatively consider a conversation flow change where you can better understand the users intention (show more or read more).

PR #328, opened 2018-11-20
Enterprise bot: Fix to prevent dropping any messages

lauren-mills:
## Description

If channel does not send conversation update messages, the first message/event will be used to trigger the start activity. If the channel does send conversation updates, the conversation update for the bot will be used to trigger the start activity. This is because there could be multiple conversationUpdates for the user(s) but only one for the bot.


darrenj:
Can you raise an issue to track the required doc changes as a result of these changes (ahead of a release)

Issue #163, opened 2018-10-22
Email template parts are hardcoded instead of localized


lzc850612:
**Description**
Email template parts are hardcoded instead of localized

In our code e.g. AI\solutions\Virtual-Assistant\src\csharp\skills\EmailSkill\Dialogs\SendEmail\SendEmailDialog.cs, we have the email template hardcoded. we should use response from localization resources. This should be fixed for all the places in the code for email templates (and other templates)

PR #341, opened 2018-11-21
Fix calendar localization issue

KayMKM:
## Description
Fix calendar localization issue

## Related Issue
Same with the issue in email skill: #163 
Some strings are hardcoded instead of localized

## Testing Steps
1. Make a zh-cn bot and check the localization string.
2. Use voice input to make sure details are all read out.


solutions/Virtual-Assistant/src/csharp/microsoft.bot.solutions/Resources/CommonStrings.de.resx
```diff
@@ -225,4 +225,37 @@
   <data name="Yesterday" xml:space="preserve">
     <value>gestern</value>
   </data>
+  <data name="AllDay" xml:space="preserve">
```
lauren-mills: Seems like these should not be in the shared library if they are related to a specific skill. They should move into a resx file in the shared folder of the skill they are used in.

lauren-mills (dismissed):

solutions/Virtual-Assistant/src/csharp/microsoft.bot.solutions/Resources/CommonStrings.de.resx
KayMKM: @lauren-mills Sure. Had moved it to calendar skill.

darrenj (approved):

PR #360, opened 2018-11-27
added EnterpriseBotSample project

lauren-mills:
## Description

Adding a sample version of the Enterprise Template so people who have already built a project with previous versions can see the latest changes easily.


ryanisgrig:
How often should this be updated? With each iterative change to the enterprise template or when there is a new release?

lauren-mills:
I think with each update. It just takes a second to recreate the project from the template, and it helps verify that it works as well.

lauren-mills:
It also will help us with bugs that happen between versions

ryanisgrig:
Makes sense!

ryanisgrig (approved):

Issue #51, opened 2018-10-10
P0: (To do) Align fully with provided LUIS model and utilise all entities to avoid prompts


lauren-mills:

bobokids:
Close this task as we already added the support to utilize the LUIS model entities. We will track issues here in specific bugs.

PR #364, opened 2018-11-28
Feich/update to do patterns in dispatch lu

feich-ms:
Update ToDo patterns in dispatch.lu
## Description
Last time, I updated the luis model in skill project and forgot to update the patterns in VA manually. This PR is to update ToDo patterns in dispatch.lu

## Related Issue
#51 

- [x] All languages have been updated
- [x] You have tested deployment with your new models


lauren-mills:
It looks like you only have two of the patterns for todo in the dispatch.lu file now... Seems like that would cause the other patterns to not be routed to the ToDo skill by the dispatcher?

feich-ms:
Actually not that. When I run the script in VA, it will pull all the patterns from todo.lu into dispatch.luis but missed the pattern.any entities. Here I only keep two patterns in dispatch.lu to let the script generate pattern.any entities in dispatch.luis. You can try the script in VA, you will find that the patterns of pointofinterest are all duplicated in the dispatch.luis.

获取 Outlook for iOS<https://aka.ms/o0ukef>

________________________________
收件人: lauren-mills <notifications@github.com>
发送时间: 星期四, 十一月 29, 2018 12:49 上午
收件人: Microsoft/AI
抄送: Fei Chen; Author
主题: Re: [Microsoft/AI] Feich/update to do patterns in dispatch lu (#364)


It looks like you only have two of the patterns for todo in the dispatch.lu file now... Seems like that would cause the other patterns to not be routed to the ToDo skill by the dispatcher?

―

feich-ms:
@lauren-mills I can give more detailed explanations about the issues in current update_all_deployment_scripts.bat. 
1. The script will pull all the pattern utterances from sub-skill lu files such as todo.lu into dispatch.luis. But the entities in the patterns are not pulled into dispatch.luis. You can remove all the patterns in the dispatch.lu to generate the dispatch.luis. You will find all the patterns from sub-skills are there but entities are not, so the luis portal will not accept this model.
2. If we copy all the pattern utterances from sub-skill lu files to the dispatch.lu, run the script and then the generated dispatch.luis will have the pattern.any entities based on the copied pattern utterances in dispatch.lu. But all the pattern utterances will have two instances in the dispatch.luis file, one is pulled from sub-skill and another one comes from the copied utterances in the dispatch.lu. This will cause the duplicated patterns in luis model file so the luis portal will not accept it either.
3. My solution here is to move two patterns with the pattern.any entities from sub-skill lu file to the dispatch.lu so that the generated dispatch.luis will have not only the patterns from sub-skills but also the needed entities with no duplicated patterns.

Please let me know if you have any questions.

lauren-mills:
Which version of the ludown tool do you have installed?

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

feich-ms:
1.0.36-84

lauren-mills:
That explains it. The way I wrote it before was assuming the public package version 1.0.36 since that is what customer’s are likely to have, and that version does not pull in the patterns from the child file into dispatch. We should stick with whatever the public packages support until the new features migrate to the npm feed so that people using the repo don’t get confused.

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

feich-ms:
All right, I will try version 1.0.36 and test the script again.

feich-ms:
@lauren-mills updated the ludown to 1.0.36 version and the ludown works now. I added all the patterns in todo.lu to the dispatch.lu. Please review again.

lauren-mills (approved):

PR #392, opened 2018-12-05
[Calendar] Join Event Dialog (Connect to Meeting)

zidaneymar:
## Description

From Gabo's sample code.
Add the join event dialog which could help extract the phone number of Teams or Skype meeting and return it as an event message to the client.
It's good to know that this dialog doesn't enable the client to join the meeting, but simply returns the phone number as an event. So it depends on the client behavior on capturing the event.

Sample utterance,
image

The phone number event:
image


solutions/Virtual-Assistant/src/csharp/skills/calendarskill/CalendarSkill.csproj
```diff
@@ -21,6 +21,7 @@
   <ItemGroup>
     <PackageReference Include="Google.Apis.Calendar.v3" Version="1.35.2.1354" />
     <PackageReference Include="Google.Apis.People.v1" Version="1.25.0.830" />
+    <PackageReference Include="HtmlAgilityPack" Version="1.8.10" />
```
darrenj, line 24: Can you test this change in skill-mode through the Virtual Assistant. I suspect the addition of this package will cause the Skill to fail as the package isn't also added to the Virtual Assitant project as well (required because of the inproc loading of DLLs)

darrenj (changes requested):
Thanks - one comment inline around package reference. Please test with the Virtual Assistant (skill mode first). Also please add a unit test to enable testing of this scenario.

solutions/Virtual-Assistant/src/csharp/skills/calendarskill/CalendarSkill.csproj
zidaneymar: image

@darrenj I clean the whole solution and start from the single virtual assistant project, and it seems that this scenarios works. Do I need to add this reference to virtual assistant as well?

I have tried to use XML parser under System.xml but it couldn't parse the some html sections correctly.

darrenj: Thanks for testing - interesting, typically the libraries need to be available but perhaps they are pulled through another package so already covered? Anyway thanks for testing.

darrenj (approved):

PR #487, opened 2018-12-21
change self begin call to replace

Theodosia1005:
## Description

Fix potential stack overflow bug when begin dialog same as the caller as we do not need the caller any more.

## Testing Steps

Create meeting
....
What time
[give a wrong time]

Check the stack.

## Checklist


lzc850612:
As i stated in the email, i believe the best way to solve this problem is to use the validator pattern. let's discuss further on how to do this.

lzc850612:
Feel free to get rid of this PR and fix this issue with validator pattern

Issue #491, opened 2018-12-21
Unable to create ToDo skill from VA


gabog:
**Project**
Virtual Assistant, Skills (ToDo)

**Description**
I am getting a "Constructor on type 'ToDoSkill.ToDoSkill' not found." exception when trying to create a ToDo item from the main VA bot.

From looking the source, it seems that @feich-ms  added MailService to the ToDoSkill constructor and the activator from SkillDialog is failing.

        public ToDoSkill(ISkillConfiguration services, ConversationState conversationState, UserState userState, IBotTelemetryClient telemetryClient, ITaskService serviceManager = null, IMailService mailService = null, bool skillMode = false)

**To Reproduce**
1. Pull the latest from master.
2. Start the emulator and say "add milk to my to do list"
You'll get a "Sorry, something went wrong." message.

**Expected behavior**
I would like to be able to add To Dos from the Virtual Assistant.

**Screenshots**
Exception in SkillDialog
![image](https://user-images.githubusercontent.com/12264946/50349423-bf124180-0509-11e9-8089-a536da551ff8.png)

Error en emulator
![image](https://user-images.githubusercontent.com/12264946/50349454-d2251180-0509-11e9-8cc3-50df4c8683e5.png)

lauren-mills:
Closed by #492

PR #494, opened 2018-12-22
[To Do] fix constructor issues in todo skills

feich-ms:
Fix constructor issues in todo skills
## Description
1. Move all the services in todo skill to service manager
2. Only service manager is kept in the constructor
3. Init all the services by service manager
3. Fix tests accordingly

## Related Issue
#491

## Testing Steps
Launch VA to test to do skill

## Checklist


solutions/Virtual-Assistant/src/csharp/skills/todoskill/Dialogs/Main/MainDialog.cs
```diff
@@ -335,6 +333,18 @@ private void InitializeConfig(ToDoSkillState state)
 
                 state.ReadSize = readSize <= 0 || readSize > CommonUtil.MaxReadSize ? CommonUtil.MaxReadSize : readSize;
             }
+
+            if (state.TaskServiceType == ProviderTypes.Other)
+            {
+                state.TaskServiceType = ProviderTypes.Outlook;
+                if (_services.Properties.ContainsKey("TaskServiceProvider"))
```
lzc850612: use _services.Properties.TryGetValue("TaskServiceProvider")

lzc850612 (changes requested):

solutions/Virtual-Assistant/src/csharp/skills/todoskill/Dialogs/Main/MainDialog.cs
feich-ms: fixed

solutions/Virtual-Assistant/src/csharp/skills/todoskill/Dialogs/Main/MainDialog.cs
```diff
@@ -305,21 +303,21 @@ private async Task<InterruptionAction> OnLogout(DialogContext dc)
...
-            // Initialize PageSize and ReadSize when the first input comes.
+            // Initialize PageSize, ReadSize and TaskServiceType when the first input comes.
             if (state.PageSize <= 0)
             {
                 int pageSize = 0;
-                if (_services.Properties.ContainsKey("DisplaySize"))
+                if (_services.Properties.TryGetValue("DisplaySize", out object displaySizeObj))
                 {
-                    pageSize = int.Parse(_services.Properties["DisplaySize"].ToString());
+                    int.TryParse(displaySizeObj.ToString(), out pageSize);
```
lzc850612, line 320: you need to handle the case where the int.TryParse fails. Log that exception and show some message to user saying its a bot issue.

lzc850612 (changes requested):

solutions/Virtual-Assistant/src/csharp/skills/todoskill/Dialogs/Main/MainDialog.cs
feich-ms: @lzc850612 Here we have default value for "PageSize"(CommonUtil.MaxDisplaySize) and "ReadSize"(CommonUtil.MaxReadSize). This is the reason why I don't want to throw exception here. The flow should not be blocked by this minor config issue. If developers see the unexpected behavior of their config setting not working, they should be able to debug and find the cause. So I prefer not to throw exception here. Please let me know if you think the exception must be thrown here. Thanks.

lzc850612: 'PageSize' is a hard dependency on the appsetting. what if this appsetting is compromised or have an illegal value? Do we just throw an unhandled exception or do we show something useful to the user and log the exception so the developer can look at it? We need to fail gracefully for all cases.

lzc850612 (approved):

Issue #508, opened 2018-12-28
Need to handle auth failure gracefully


lzc850612:
**Project**
Virtual Assistant

**Description**
There's intermittent issue when trying to get authenticated for auth and we're not handling the exception case gracefully. 

**To Reproduce**
Sometimes when finishing an auth flow after inputting the magic code, the token response from ABS is null and we'll see exceptions that's handled

**Expected behavior**
We should handle the exception and send back a user-friendly message

**Screenshots**
![image](https://user-images.githubusercontent.com/4001684/50529452-bdb4bc00-0aa9-11e9-9c75-691851ef1fe8.png)

PR #509, opened 2018-12-28
Add exception handling for auth

lzc850612:
## Description
When the OAuth card comes back with empty response, we need to handle that case. Right now we're not so we'll see exceptions when that happens. This change is to handle that case gracefully

## Related Issue
#508 

## Testing Steps
Open Emulator with a new user id and send 'send an email' to the bot to trigger the OAuth card. Sometimes after inputting the magic code, there will be an error.

## Checklist


solutions/Virtual-Assistant/src/csharp/microsoft.bot.solutions/Authentication/MultiProviderAuthDialog.cs
```diff
@@ -119,12 +121,12 @@ private async Task<DialogTurnResult> PromptForAuth(WaterfallStepContext stepCont
             var authType = string.Empty;
```
DingmaomaoBJTU: >var authType = string.Empty; [](start = 12, length = 28)

Guess we don't need this? Just use _selectedAuthType instead?

DingmaomaoBJTU (approved):
:shipit:

solutions/Virtual-Assistant/src/csharp/microsoft.bot.solutions/Authentication/MultiProviderAuthDialog.cs
lzc850612: yes you're right.

Issue #512, opened 2018-12-31
Duplicate code when trying to log stuff


lzc850612:
**Project**
Virtual Assistant

**Description**
We have duplicate code when trying to log with a common set of properties.

lauren-mills:
Closed by #513

Issue #511, opened 2018-12-31
ISkillConfiguration is a wrong name for an abstract class


lzc850612:
**Project**
Virtual Assistant

**Description**
ISkillConfiguration is an abstract but named like an interface

lauren-mills:
Closed by #513

Issue #510, opened 2018-12-31
Using Customized code for Telemetry instead of the native one from 4.2 SDK for skills


lzc850612:
**Project**
Virtual Assistant

**Description**
In skills we are still using the TelemetryClient passed from VA that's initialized from customized code instead of native support from 4.2 sdk.

lauren-mills:
Closed by #513

Issue #394, opened 2018-12-05
Skills Template


darrenj:
**Project**
Virtual Assistant

**Description**
The first pass at a Skill Template to help our internal work needs formalising into a formal Skill Template to allow someone to create a new Skill and enable an "F5" experience to have this working out of the box so they can plug into the VA without having to worry about Skill specifics enabling them to focus on their own skill code.

- [ ] Ensure project is .NET Core 2.2 (or latest) out of the box
- [x] Rationalise ISkillConfiguration and SkillConfiguration on constructor and ensure template represents our preferred direction. Some of our skills seems to use ISkillConfiguration now (so we should align too). Also requires changes to Startup.cs
```
var supportedProviders = Configuration.GetSection("SupportedProviders")?.Get<string[]>();
ISkillConfiguration connectedServices = new SkillConfiguration(botConfig, supportedProviders, parameters, configuration);
services.AddSingleton(sp => connectedServices);
```
- [x] Update to new DeploymentScripts approach
- [x] CognitiveModels/LUIS/en folder not created, perhaps add a skill.lu file placeholder?
- [x] Add Actions.cs file to Shared folder
- [x] Starter LU file for a ExampleDialog and wire intent into RouteAsync/MainDialog
- [x] Starter Dialog which is wired up ready to go - e.g. ExampleDialog
- [x] Update to reflect new auth approach for multiple providers
- [x] Remove News Nuget package reference
- [x] Template created csproj has lamillsbot-insights references which need to be removed
- [x] OnLogout handler in MainDialog refers to old AuthConnectionName
- [x] {SkillName}Dialog constructor has old oAuthPrompt/AuthConnectionName code
- [x] Add stylecop analyzer nuget
- [x] Ensure the MainResponses have the right InputHint - e.g. acceptingInput rather than expectingInput for greeting
- [x] End to End Test
- [x] VSIX
- [x] Documentation guiding the user on how to add a LUIS model and Dialog and wire up
- [x] Documentation guiding the user on how to take a skill and add to VA
- [x] Skill development guide (best practice)

lauren-mills:
I'm removing the 2.2 dependency because our build pipeline doesn't support it yet. Once the build and release pipelines support 2.2 we will update all the projects at the same time.

PR #513, opened 2018-12-31
revamp telemetry client & skill configuration

lzc850612:
## Description
There's duplicate code when trying to log stuff for same/similar set of properties across the solution. Need to implement extension methods to handle the basic properties to avoid duplicate code.

We're using 4.2 bot builder sdk now and the TelemetryClient comes with native support, which means we don't have to setup the TelemetryClient ourselves. In the skills though we're still using the TelemetryClient we load using customized code. We should switch over to the native TelemetryClient altogether.

ISkillConfiguration is also wrongly named like an interface so changing it to a proper abstract class name SkillConfigurationBase and all the references

## Related Issue
#510 
#511 
#512 

## Checklist


darrenj:
We should also update the experimental skills for consistency and add additional steps to #394

lzc850612:
> We should also update the experimental skills for consistency and add additional steps to #394

Done

lauren-mills (approved):

darrenj (approved):

PR #532, opened 2019-01-10
add date prompt to create

Theodosia1005:
## Description
Change the Create Get Date to Date Prompt

## Testing Steps
test in the date part with 
  alalalal (not datetime)
  afternoon (not date)
  skip (set default as today)

## Checklist

### DIFF
--- a/solutions/Virtual-Assistant/src/csharp/skills/calendarskill/Dialogs/CreateEvent/DatePrompt.cs
+++ b/solutions/Virtual-Assistant/src/csharp/skills/calendarskill/Dialogs/CreateEvent/DatePrompt.cs
@@ -1,0 +1,156 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using CalendarSkill.Dialogs.CreateEvent.Resources;
+using Microsoft.Bot.Builder;
+using Microsoft.Bot.Builder.Dialogs;
+using Microsoft.Bot.Schema;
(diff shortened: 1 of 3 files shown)


solutions/Virtual-Assistant/src/csharp/skills/calendarskill/Dialogs/CreateEvent/DatePrompt.cs
```diff
@@ -0,0 +1,156 @@
+﻿using System;
```
lzc850612, line 1: I feel that we should move this class to Microsoft.Bot.Solutions since this is a generic date prompt implementation.

lzc850612 (changes requested):

solutions/Virtual-Assistant/src/csharp/skills/calendarskill/Dialogs/CreateEvent/DatePrompt.cs
Theodosia1005: it has some specific logic for create flow now. Like skip, like allow it to have date and time at the same time. We need to move skip to Bot Solution too.

lzc850612: ok let's keep this in here for now and if other skills are starting to use this too we'll think about the right level of abstraction of putting this to microsoft.bot.solutions.

lzc850612 (approved):

PR #543, opened 2019-01-13
Added Speak variants for speech scenarios in news experimental skill

darrenj:
## Description
Populating the Speak property on the news article response so it behaves well in speech scenarios. Use the description field (albeit truncated) to provide some summary.

### DIFF
--- a/solutions/Virtual-Assistant/src/csharp/experimental/skills/newsskill/Dialogs/FindArticles/FindArticlesResponses.cs
+++ b/solutions/Virtual-Assistant/src/csharp/experimental/skills/newsskill/Dialogs/FindArticles/FindArticlesResponses.cs
@@ -38,6 +38,16 @@ namespace NewsSkill
             if (articles.Any())
             {
                 response.Text = "Here's what I found:";
+
+                if (articles.Count > 1)
+                {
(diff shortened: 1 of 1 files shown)


solutions/Virtual-Assistant/src/csharp/experimental/skills/newsskill/Dialogs/FindArticles/FindArticlesResponses.cs
```diff
@@ -38,6 +38,16 @@ private static object ShowArticleCards(ITurnContext context, dynamic data)
             if (articles.Any())
             {
                 response.Text = "Here's what I found:";
+
+                if (articles.Count > 1)
+                {
```
lzc850612, line 43: so it's better to put these strings to the *.json file for different languages. I'm assuming we'll do this when we move this skill to the Virtual Assistant solution out of the 'experimental' realm

darrenj: That was my thinking, I can go and do this if we think experimental should follow the pattern from the beginning

PR #628, opened 2019-01-25
[Automotive] Add local deployment files to .gitignore

outofthecave:
## Description

These files are automatically generated when deploying the bot locally and therefore do not need to be tracked by git.

## Related Issue

I don't think we need an issue for this, but if you'd prefer me to create one, just let me know and I'll do so.

## Testing Steps

This does not apply here.

## Checklist
This does not apply here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -340,3 +340,8 @@ ASALocalRun/
 
 # QnAMaker json files
 /solutions/**/*.qna
+
+# Local deployment files
+/solutions/**/.deployment
+/solutions/**/publish
+/solutions/**/publish.cmd


.gitignore
```diff
@@ -339,4 +339,9 @@ ASALocalRun/
 /solutions/**/*.luis
 
 # QnAMaker json files
-/solutions/**/*.qna
\ No newline at end of file
+/solutions/**/*.qna
+
+# Automotive skill local deployment files
+/solutions/Virtual-Assistant/src/csharp/skills/automotiveskill/automotiveskill/.deployment
```
lauren-mills: This is a good call, but it'd be better to use a path with wildcards so all these files are ignored.

lzc850612: yes very good suggestion

lzc850612 (changes requested):

.gitignore
outofthecave: Thanks!  I changed it to `/solutions/Virtual-Assistant/src/csharp/skills/*skill/*skill/.deployment` - I hope that's what you meant.  If not, let me know.

lauren-mills (approved):

outofthecave:
@lzc850612 I made the changes you requested.  Can you have another look please?

lzc850612:
I think it should be good enough to do 
/solutions/**/.deployment
/solutions/**/publish.cmd
/solutions/**/publish

They would be generated when we publish to azure and that's true for all web apps, not just skills.

Thanks!

PR #629, opened 2019-01-25
Fix for deploying the bot to an existing resource group

gabog:
## Description
Updated first call to msbot clone to use the groupName parameter.
Updated the second call to use $groupName instead of $name

## Related Issue
Using deploy_bot.ps1 with -groupName doesn't use the resource group name when executing msbot clone.

## Testing Steps
Precreate a resource group in azure with any name, i.e. TestResourceGroup
execute
debloy_bot.ps1 -name MyTestBot -groupName TestResourceGroup
The bot should be created in the provided resource group (before it as created a resource group with the bot name and creating the resources there rather than using the resource group parameter)


lauren-mills (changes requested):
Does this make the groupName required? Ideally, it'd be optional, and set to the name if not supplied

gabog:
You are right, I'll fix it.
Thanks

gabog:
@lauren-mills, I added a default to the parameter, let me know what you thing.

Sorry it took so long.

lauren-mills (approved):

PR #637, opened 2019-01-28
Automotive Skill updates for Speech Support, Images and VA Dispatch integration

darrenj:
## Description
- Speak property now set for a higher quality speech experience
- New images for setting validation cards
- Cut-down LU model for use by dispatch
- Updated skill deployment with steps required to add VA **prior** to deployment
- Skill Mode required a different approach for resolving image resources than the existing "local-mode" which can use HttpContext meaning images "just work" in local-mode. Developer will need to ensure image assets are on CDN/HTTP location and provide location via skillconfiguratuion entry.
- Updated Skill constructor signature to move ServiceManager (not used by Skill invocation) to end and move skillMode from end to align better.

## Testing Steps
- Run Unit Tests to validate internal skill changes
- Follow `Deploying the Virtual Assistant with the Automotive Skill` in docs to configure VA skill ahead of VA deployment

## Checklist
- [x ] I have added or updated the appropriate unit tests

- [ ] All languages have been updated   (No localisation in this skill at this time)


solutions/Virtual-Assistant/src/csharp/skills/automotiveskill/automotiveskill/Dialogs/VehicleSettings/VehicleSettingsDialog.cs
```diff
@@ -271,6 +281,25 @@ private async Task<DialogTurnResult> ProcessVehicleSettingsChange(WaterfallStepC
                     if (settingValues.Count() > 1)
                     {
                         string settingName = state.Changes.First().SettingName;
+                        var setting = this.settingList.FindSetting(settingName);
+                        var categoryName = setting.Categories.First();
+
+                        string imageName;
+                        switch (categoryName)
+                        {
+                            case "Climate Control":
```
outofthecave: I was actually planning to remove the category names altogether because they're somewhat specific to each customer and they're only used in one place (in the confirmation prompt) and I was planning to change that too because that's also customer-specific.  So, I'd suggest that instead of relying on the category names here, we could add a specific property to `AvailableSetting` for the image file name for each setting.  That way, it would be easier for less technical bot developers to customize the image by editing the YAML or JSON files that list all the settings.

darrenj: That would be much better, that's how I thought we would go once we did the "generalisation" of the JSON/YAML files. Shall we go with this for now and then fix-up in your PR? Or happy the other way round if you prefer

outofthecave: I'd find it better to keep the two PR's independent of each other and to add the new property for the image file name in this PR and only remove the categories in my PR.  Does that make sense?

darrenj: Yes, makes sense. Happy to make that change right now. As of now a AvailableSetting can be part of multiple Categories. Is that actually the case or can we move this to Setting==Category? If so I can add CategoryImageFileName property to AvailableSetting.cs otherwise I'll need to add a new Category class and add a collection to AvailableSetting. (Right now we only ever pick the first category so ignore anything else)

darrenj:
@lauren-mills can you do a quick sanity check of the skilldialog changes, they'll need to be a tweak to skill template too

solutions/Virtual-Assistant/src/csharp/skills/automotiveskill/automotiveskill/AutomotiveSkill.csproj
```diff
@@ -6,6 +6,60 @@
     <CodeAnalysisRuleSet>..\..\..\VirtualAssistant.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
 
+  <ItemGroup>
```
lauren-mills: I think these can be removed

solutions/Virtual-Assistant/docs/virtualassistant-skills-automotive.md
```diff
@@ -202,4 +204,47 @@ msbot list --bot YOURBOTFILE.bot --secret YOUR_BOT_SECRET
...
 ## Testing the skill in local-mode
 
-Once you have followed the deployment instructions above, open the provided .bot file with the Bot Framework Emulator.
\ No newline at end of file
+Once you have followed the deployment instructions above, open the provided .bot file with the Bot Framework Emulator.
+
+## Deploying the Virtual Assistant with the Automotive Skill
+
+1. Update the Virtual Assistant deployment scripts.
+    - Add the addition automotive skill LUIS models to the bot.recipe file located witnin your assistant project: `assistant\DeploymentScripts\en\bot.recipe`
```
lauren-mills, line 212: *additional

lauren-mills (approved):

solutions/Virtual-Assistant/src/csharp/skills/automotiveskill/automotiveskill/Dialogs/VehicleSettings/VehicleSettingsDialog.cs
outofthecave: I actually meant we should get rid of the categories completely.  Not all car manufacturers divide their settings into categories, so we shouldn't assume that.  Instead, each setting can just have a property for the image file to use for that setting (without any reference to categories).  If a manufacturer does have a notion of categories and if the images are based on those categories, they can just use the same image path for all the settings that form a category.  But our code doesn't need to know anything about the categories.

solutions/Virtual-Assistant/src/csharp/skills/automotiveskill/automotiveskill/AutomotiveSkill.csproj
```diff
@@ -5,6 +5,45 @@
...
+    <Content Include="Dialogs\VehicleSettings\Resources\normalization\amount_percentage.tsv">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
+    <Content Include="Dialogs\VehicleSettings\Resources\normalization\amount_type.tsv">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
+    <Content Include="Dialogs\VehicleSettings\Resources\normalization\amount_unit.tsv">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
+    <Content Include="Dialogs\VehicleSettings\Resources\normalization\color.tsv">
```
outofthecave, line 19: [This comment shouldn't stop this PR from being merged.]
Btw, most of these files aren't even used for vehicle settings.  They're from a different skill.  Vehicle settings only needs the ones that start with `amount` and `index_map.tsv`.

solutions/Virtual-Assistant/src/csharp/skills/automotiveskill/automotiveskill/Dialogs/VehicleSettings/Resources/available_settings.json
```diff
@@ -1,7 +1,7 @@
 [
   {
     "canonicalName": "Front Left Air Delivery Mode Control",
-    "categories": [ "Climate Control" ],
+    "category": "Climate Control",
```
outofthecave, line 4: If you're changing this file, I'd recommend to either change the corresponding YAML file to keep them in sync or delete the YAML file to avoid confusion.

outofthecave:
@darrenj Sorry, I didn't realize that the PR was already merged when I posted my comments.  I guess they're moot then.  I'll just make another PR for the changes I suggested, then we can discuss there.

solutions/Virtual-Assistant/src/csharp/skills/automotiveskill/automotiveskill/AutomotiveSkill.csproj
darrenj: Thanks - will get those removed for simplicity.

solutions/Virtual-Assistant/src/csharp/skills/automotiveskill/automotiveskill/Dialogs/VehicleSettings/Resources/available_settings.json
darrenj: I was going to ask if we wanted to maintain YAML alongside JSON, is there a tool/script you use for converting?

darrenj:
As I merged it came through - all good, I get them incorporated (see above)

solutions/Virtual-Assistant/src/csharp/skills/automotiveskill/automotiveskill/Dialogs/VehicleSettings/Resources/available_settings.json
outofthecave: Well, I wrote a 19-line Python script to convert it: https://vsogd.visualstudio.com/_git/GM%20IPA%20PoC?path=%2FSrc%2FIPABot%2FDialogs%2FVehicle%2Fyaml_to_json.py&version=GBvehicle-settings-original, but any converter will do.  I tried to preserve the order of the properties in each JSON object, to make it easier to read for people who are used to our YAML files (i.e. mostly myself).

darrenj:
OK – feel free to add to the repo with your PR. Thanks!

PR #644, opened 2019-01-28
Properly end single turn dialogs

mekinney:
Single turn dialogs currently don't properly end the active dialog.  For these dialog no DialogTurnStatus complete is ever sent.  Only a single Empty is sent.  So check if that is the case and then end the dialog.

Check for empty dialog stack in the dialog turn status empty case.  Perform proper end dialog processing in this case.

## Related Issue
https://github.com/Microsoft/AI/issues/643

## Testing Steps
Create a simple single turn dialog (code in issue).

### DIFF
--- a/templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Shared/RouterDialog.cs
+++ b/templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Shared/RouterDialog.cs
@@ -48,13 +48,9 @@ namespace $safeprojectname$.Dialogs.Shared
                                 case DialogTurnStatus.Empty:
                                     {
                                         await RouteAsync(innerDc);
-                                        // Waterfalls with no turns should fire EndDialog code.
+                                        // Waterfalls with no turns should Complete.
                                         if (innerDc.Stack.Count == 0)
-                                        {
                                             await CompleteAsync(innerDc);
-                                            // End active dialog
-                                            await innerDc.EndDialogAsync();
-                                        }
                                         break;
                                     }
 
--- a/templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Shared/RouterDialog.cs
+++ b/templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Shared/RouterDialog.cs
@@ -48,6 +48,13 @@ namespace $safeprojectname$.Dialogs.Shared
                                 case DialogTurnStatus.Empty:
                                     {
                                         await RouteAsync(innerDc);
+                                        // Waterfalls with no turns should fire EndDialog code.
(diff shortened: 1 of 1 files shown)


templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Shared/RouterDialog.cs
```diff
@@ -48,6 +48,13 @@ protected override async Task<DialogTurnResult> OnContinueDialogAsync(DialogCont
                                 case DialogTurnStatus.Empty:
                                     {
                                         await RouteAsync(innerDc);
+                                        // Waterfalls with no turns should fire EndDialog code.
+                                        if (innerDc.Stack.Count == 0)
+                                        {
+                                            await CompleteAsync(innerDc);
+                                            // End active dialog
+                                            await innerDc.EndDialogAsync();
```
lauren-mills: This line shouldn't be necessary, because there is nothing on the stack to end.

lauren-mills (changes requested):

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Shared/RouterDialog.cs
mekinney: I'm not sure which line you mean by "this" line.  Are you referencing the entire commit?

lauren-mills: Sorry, the `await innerDc.EndDialogAsync()` line.

Since you're checking if the stack is empty before getting here, there won't be anything in the innerDc to end.

mekinney: I agree with you, however... :)

The DialogTurnStatus.Complete case performs EndDialogAsync regularly when their are no dialogs on the stack.  Furthermore, for my cases, DialogLogTurnStatus.Complete only occurs when there are no dialogs on the stack.  I was assuming EndDialogAsync was in the Complete case for a reason, so I duplicated it in the one turn case.

That said, I'm happy to change the commit to omit it, but wanted to point out the inconsistency so that someone with more knowledge of the code could consider it.

mekinney: Updated the commit.

lauren-mills (approved):

templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/Bot Framework/Dialogs/Shared/RouterDialog.cs
lauren-mills: Noted. I'll check it out

PR #662, opened 2019-01-30
Telemetry: QnA does not need to add activityID

daveta:
## Description
Fixes collision in Application Insights where QnA recognizer adding property that the initializer automatically puts in..

This needs to be replicated to VA (and other locations in Enterprise (?)).


lzc850612 (approved):

lzc850612:
Thanks Dave. I'll make these changes in VA as well.

PR #692, opened 2019-02-01
Ensure nested Enterprise Dialogs do not throw error adding luis result to TurnState

garypretty:
## Description
Fix to ensure nested enterprise dialogs do not throw error trying to add LUIS result to TurnState twice and also ensure LUIS is only called once per turn.


garypretty:
Not sure why this diff looks so messed up.  Essentially the code I have changed boils down to the following check to see if we already have a LUIS result before calling the recognizer and adding it to the TurnState. cc @darrenj 

```cs
 General luisResult;
            if (dc.Context.TurnState.ContainsKey(LuisResultKey))
            {
                luisResult = dc.Context.TurnState.Get<General>(LuisResultKey);
            }
            else
            {
                luisResult = await luisService.RecognizeAsync<General>(dc.Context, true, cancellationToken);

                // Add the luis result (intent and entities) for further processing in the derived dialog
                dc.Context.TurnState.Add(LuisResultKey, luisResult);
            }
```

lzc850612 (approved):

PR #773, opened 2019-02-13
Updated WebTest pages so they can be used with unified speech

gabog:
## Description
Fixes to the web test project so it can be used with Unified speech.
Added a SpeechRegion config setting and updated WebChat.cshtml to use unified speech.
If SpeechKey and SpeechRegion (westus, eastus, etc.) are provided in app settings then the microphone will render and you will be able to use speech with webchat.
Note: the latest webchat does not support changing the voice font yet, I left the parameter in config for future use. 

## Related Issue
Need to be able to test the assistant on WebChat with speech.

## Testing Steps
1. Create a Speech service in azure.
2. Update appsettings.json with the service key and region.
3. Add the other parameters required by the project (auth keys, directline, etc.)
4. Run the project and you should be able to use speech.

### DIFF
--- a/solutions/Virtual-Assistant/src/csharp/testharnesses/assistant-WebTest/Assistant-WebTest/Controllers/HomeController.cs
+++ b/solutions/Virtual-Assistant/src/csharp/testharnesses/assistant-WebTest/Assistant-WebTest/Controllers/HomeController.cs
@@ -2,36 +2,46 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Collections.Generic;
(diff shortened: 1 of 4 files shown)


gabog:
This PR also fixes linked accounts, I renamed the **Bot**MicrosoftAppId and **Bot**MicrosoftAppPassword to MicrosoftAppId and MicrosoftAppPassword so the OOTB CredentialsProvider can find the settings and doesn't crash.

ryanisgrig (approved):
Validated and tested deployed version - thanks!

Issue #790, opened 2019-02-14
Deployment Issues when providing appSecret with special characters


martinoss:
**Project**
Virtual Assistant

**Description**
Having special characters in the appSecret causes an error when calling msbot internally. The script continues to run but only deploys the language specific resources.

**To Reproduce**
Deploy with problematic app secret:
`PowerShell.exe -ExecutionPolicy Bypass -File DeploymentScripts\deploy_bot.ps1 --appId "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa" --appSecret "5xUzz-I])4(M!LS}73xpTU%n"`

**Expected behavior**
The script `deploy_bot.ps1` should escape the secret param. 

**Screenshots**

![image](https://user-images.githubusercontent.com/2587942/52794721-ad17cf80-3070-11e9-869b-977ccdcc69b5.png)

**Additional context**

martinoss:
see also https://github.com/Microsoft/botbuilder-tools/issues/694#issuecomment-460238166

PR #791, opened 2019-02-14
Escape appSecret when passing to msbot clone

martinoss:
## Description
Add `--%` as [documented on this page](https://github.com/Microsoft/botbuilder-tools/blob/master/packages/MSBot/docs/export-clone.md) for powershell calls 

## Related Issue
#790 

## Testing Steps
Start the deployment of the virtual assistant using following appSecret. It will output an error without the fix.
`PowerShell.exe -ExecutionPolicy Bypass -File DeploymentScripts\deploy_bot.ps1 --appId "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa" --appSecret "5xUzz-I])4(M!LS}73xpTU%n"`


martinoss:
Oh wait - please don't merge it yet... It works to deploy but the appSecret will be $appSecret

Issue #103, opened 2018-10-15
P0: Proactive Messaging Support


lzc850612:
We need to implement proactive conversation such as sending an message proactively to the user to say e.g. 'You have a meeting in an hour'.  We need to implement this for different skills based on events and predefined criteria

Issue #66, opened 2018-10-10
P1: (Calendar) Proactive "Upcoming meeting"


lauren-mills:

lzc850612:
Won't be fully supported until Vanadium. Will implement the dialog for Upcoming Meeting but not the webjob processor.

darrenj:
tracked seperately

PR #800, opened 2019-02-15
first proactive scenario

lzc850612:
## Description
This change is to enable the first proactive scenario, which is when car sends a CarStart event the bot starts checking if there's a meeting coming up in the next 60 minutes.

## Related Issue

#66 
#103 

## Testing Steps
after building and running the bot, starts with 'what's my meeting today' after finishing the dialog, send 
/event:{ "Name": "CarStart" } 
event to the bot, the bot will start checking for any meeting in the next 60 minutes for 30 minutes. then go to your outlook email account to create a meeting with someone within the next 60 minutes, then you'll see the meeting notification appearing in the chat window

## Checklist
- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] I have added or updated the appropriate unit tests
- [ x ] I have tested any new/updated dialogs using Speech in the emulator to ensure the [Speak](https://docs.microsoft.com/en-us/azure/bot-service/dotnet/bot-builder-dotnet-text-to-speech?view=azure-bot-service-3.0) property is set to enable a high quality speech-first experience and the appropriate [InputHints](https://docs.microsoft.com/en-us/azure/bot-service/dotnet/bot-builder-dotnet-add-input-hints?view=azure-bot-service-3.0) are set correctly. 

### DIFF
--- a/solutions/Virtual-Assistant/src/csharp/Microsoft.Bot.Solutions/Middleware/ProactiveStateMiddleware.cs
+++ b/solutions/Virtual-Assistant/src/csharp/Microsoft.Bot.Solutions/Middleware/ProactiveStateMiddleware.cs
@@ -2,10 +2,13 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Security.Cryptography;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Bot.Builder;
 using Microsoft.Bot.Solutions.Models.Proactive;
+using Utilities;
 using static Microsoft.Bot.Solutions.Models.Proactive.ProactiveModel;
 
 namespace Microsoft.Bot.Solutions.Middleware
@@ -33,11 +36,11 @@ namespace Microsoft.Bot.Solutions.Middleware
             {
                 var proactiveState = await _proactiveStateAccessor.GetAsync(turnContext, () => new ProactiveModel());
                 ProactiveData data;
-                var userId = turnContext.Activity.From.Id;
+                var hashedUserId = MD5Util.ComputeHash(turnContext.Activity.From.Id);
                 var conversationReference = turnContext.Activity.GetConversationReference();
                 var proactiveData = new ProactiveData { Conversation = conversationReference };
 
-                if (proactiveState.TryGetValue(userId, out data))
+                if (proactiveState.TryGetValue(hashedUserId, out data))
                 {
                     data.Conversation = conversationReference;
                 }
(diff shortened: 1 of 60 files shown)


solutions/Virtual-Assistant/src/csharp/Microsoft.Bot.Solutions/Middleware/ProactiveStateMiddleware.cs
```diff
@@ -26,24 +30,29 @@ public ProactiveStateMiddleware(ProactiveState proactiveState)
...
         {
-            var proactiveState = await _proactiveStateAccessor.GetAsync(turnContext, () => new ProactiveModel());
-            ProactiveData data;
-            var userId = turnContext.Activity.From.Id;
-            var conversationReference = turnContext.Activity.GetConversationReference();
-            var proactiveData = new ProactiveData { Conversation = conversationReference };
+            var activity = turnContext.Activity;
 
-            if (proactiveState.TryGetValue(userId, out data))
+            if (activity.From.Properties["role"].ToString().Equals("user", StringComparison.InvariantCultureIgnoreCase))
```
lauren-mills, line 35: Should be able to simplify this using: `activity.From.Role == RoleTypes.User`

solutions/Virtual-Assistant/src/csharp/utilities/TaskExtensions/BackgroundTaskQueue.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
```
lauren-mills, line 1: Do these classes need to be in a separate project for the background tasks to work correctly, or is this something we can add to the microsoft.bot.solutions project?

lauren-mills (changes requested):

solutions/Virtual-Assistant/src/csharp/Microsoft.Bot.Solutions/Middleware/ProactiveStateMiddleware.cs
lzc850612: actually From.Role is empty even though properties["role"] is user. that's why i do it this way

solutions/Virtual-Assistant/src/csharp/utilities/TaskExtensions/BackgroundTaskQueue.cs
lzc850612: So I think given we want to go into the direction of having icrosoft.bot.solutions as a nuget package, i want to have something like this that's not always being used by all skills first in a separate project. later on if we think we should add this into the microsoft.bot.solutions project, we can do that. right now i don't feel the need.

lauren-mills: Ok, that's fine for now. But I think since its enabling a type of scenario (and is not specific to any one skill) it could still safely go in the shared project in the future.

lauren-mills (approved):

PR #784, opened 2019-02-14
Telemetry middleware updates for 4.3

ryanisgrig:
## Description

This contains all updates to the middleware for next release:
* **TelemetryLoggerMiddleware**
    * Added **ITelemetryLoggerMiddleware** to allow users the ability to extend their own 
    * Added `speak` property to all events
* **TelemetryExtensions**
     * Remove duplicate properties that are provided by the SDK telemetry
* **QnaTelemetryMaker**
    *  Added `questionId` property
    * Fixed bug with `answer`
* **TelemetryLuisRecognizer**
    * Added `entities` and `fromId` properties
    * Renamed `luisIntent.IntentName` to `LuisResult`
* Updated **Microsoft.Bot.Builder** references to `4.2.2`
* Updated **SkillName.OnTurnAsync** function in all skills to pass the telemetry client, this was blocking the analytics from logging sub-intents. Presently it is only logging the Dispatch result.
* Refactored **IQnaTelemetryMaker** & **ITelemetryLuisRecognizer** to use boolean `LogPersonalInformation` to configure what PII data should be logged

## Related Issue
Close #729 
Close #730 

## Checklist
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have updated related documentation

PR #825, opened 2019-02-19
[Skills] Fix crash issue for "An item with the same key has already been added."

DingmaomaoBJTU:
## Description

[Skills] Fix crash issue for "An item with the same key has already been added."

## Checklist


ryanisgrig (approved):
Do you have any idea what property is still throwing the same key error?
I removed conflicting ones in #784

DingmaomaoBJTU:
Let me check and update to you if I find anything,

DingmaomaoBJTU:
@ryanlengel I guess the root cause is  because OnTurnAsync() will be entered multiple times during conversation, so TelemetryLoggerMiddleware.AppInsightsServiceKey will be added multiple times with the same key. So using tryadd will be good to avoid this problem~

PR #833, opened 2019-02-20
[Calendar] Update Calendar Luis

Theodosia1005:
## Description
update calendar luis

## Checklist

- [x] All languages have been updated


solutions/Virtual-Assistant/src/csharp/skills/calendarskill/calendarskill/CognitiveModels/LUIS/en/calendar.lu
```diff
@@ -103,138 +107,174 @@
...
-- will {ContactName=karen} be free on {FromDate=friday}
+- when is aaron free {FromDate=tomorrow}
+- when is bryan free
+- when is dylan available
+- when is samuel available {FromDate=tomorrow}
+- when will jordan be free
+- will tyler be free on {FromDate=friday}
+
+
+## Confirm
```
lauren-mills: This intent shouldn't be added to the calendar.lu, you should use the one in the general model instead to avoid conflicts. This applies to the showNext and showPrevious intents

solutions/Virtual-Assistant/src/csharp/skills/calendarskill/calendarskill/CognitiveModels/LUIS/en/calendar.lu
```diff
@@ -958,57 +1098,99 @@
...
+- hello world
+- how's the weather today
+- what can you do
+- what is the time now
+- what's my name
+- what's the weather today
+- what's your name
+
+
+## Reject
```
lauren-mills: This one is also a duplicate of the general model

lauren-mills (changes requested):

solutions/Virtual-Assistant/src/csharp/skills/calendarskill/calendarskill/CognitiveModels/LUIS/en/calendar.lu
Theodosia1005: But the problem is, we should not add anything about skills in general model. Or, if the general model have something like 
- Yes, please send this email
- Yes, create the meeting.
- Yes, add the todo please.
It may confuse when dispatch. For example, what if the dispatch Please send an email to General model?
For now, all the duplicate intents are not Start-Flow-Intent. So it will not make any confuse on VA dispatch part. And within the skill model, it will and should be able to get clear on send an email and Yes please send the email. The skill flow can merge the result from general and skill model when expecting the reply a confirm.

lauren-mills: What we discussed in the sync was it’s okay to have this intent as long as there are no overlapping utterances with general and the intent names are different. So all the calendar specific utterances can be in this file, but the more generic utterances like “cancel this” or “next” should be in general. That will prevent any overlap.

The dispatch approach will soon be updated to use the dispatch cli tool instead of the dispatch.lu approach we use today and I don’t think we can distinguish between the triggering intents and the intents for 2nd turn responses.

lauren-mills (approved):

PR #860, opened 2019-02-23
Remove platform dependent tools

### LINKED ISSUE #859: Virtual Assistant test and deploy scripts only run on Windows
**Project**
Virtual Assistant

**Description**
.NET Core and PowerShell both have good cross-platform support, however this project has a few minor instances where a Windows environment is assumed. Specifically:
* `AutomotiveSkillTest.csproj` uses `xcopy`in a PostBuild exec.
* All of the PowerShell scripts hardcode the use of `\` as a path separator, instead of using `Join-Path`
* the `INTRO_PATH` resource string is hardcoded to use `\` as a path separator which fails at runtime

**To Reproduce**
1. Get a Mac or Linux host that is running .NET Core 2.2 with PowerShell installed
1. Try to build the solution -> fails because `xcopy` does not exist
1. Try to run `deploy_bot.ps1` -> fails because path separators are not correct for the platform
1. Start the bot after successful build and connect a client -> fails because `INTRO_PATH` json cannot be found

**Expected behavior**
* The solution should build on all .NET supported platforms
* Supporting scripts should run anywhere that PowerShell can run
* The bot should be able to load supporting files on all .NET supported platforms

**Screenshots**
N/A

**Additional context**
I'll submit a PR shortly that addresses these

ctstone:
## Description

Fixes #859. Test and deploy scripts currently require Windows-specific tools, when cross-platform options are available. This PR does two things:
1. Test for presence of $BASH during build, and use `cp` instead of `xcopy` if available
1. Switch all path strings in `*.ps1` to use `Join-Path` instead of hardcoding `\`
1. Replace `\` in the `INTRO_PATH` string before the file is loaded from disk

### Architecture
N/A

### Bug Fixes
#859 

### Conversational Experience
N/A

### Maintenance
N/A

### Language Understanding
N/A

## Testing Steps
1. Should be tested on Windows, Mac, and Linux
1. Build should complete without errors
1. Running `deploy_bot.ps1` with all supported parameter sets should complete without errors

## Checklist
N/A


darrenj:
@lauren-mills can you run a quick sanity check on a windows box ahead of merging (in case it makes it into the release). Looks OK to me

lauren-mills (approved):

lauren-mills:
@ctstone looks good. If you can resolve the conflict I'll merge it in

ctstone:
@lauren-mills it looks like the `xcopy` fix I introduced is no longer necessary since the `Exec` build task has since been removed. I think this latest merge should close the loop

gabog:
@ctstone & @lauren-mills, I just pulled the latest and update_published_models.ps1 doesn't work for me on windows now, do you guys see the same error? 
I tried running .\update_published_models.ps1 "en-us" and I get:
image

It seems to be an issue with JoinPath, was this changed so it can work on a Mac?
image

ctstone:
@gabog, yes this change was introduced to support .NET platforms that use `/` as a path separator (namely Mac and Linux).

It turns out `Join-Path` has different behavior between PS v5 and v6 (surprising for a seemingly simple utility!): https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/join-path?view=powershell-6#optional-parameters

I verified that I see the same error in PS v5, so I'll post back with a branch that supports the older shell.

gabog:
Thanks @ctstone, I installed PS 6 Core on my machine and it works after that but I get other errors later:

image

I am not doing a deploy, just an updated on a version that I had from a week and a half ago. I am not sure if this second error is related to changes in the scripting, powershell functions or changes to the folder structure. 

Can you let me know if you see something similar when running update published models in your windows environment?

Thanks

gabog:
also, I didn't know about this but it seems we are changing the direction for PowerShell (some info here: https://redmondmag.com/articles/2017/07/17/windows-powershell-6.0-and-powershell-core.aspx). 
@darrenj, if we require PS 6 Core for cross plat we should add that to the pre-reqs list, by default, Windows Devs will use PS Desktop that comes with Windows and that one is v5.

ctstone:
TBH `Join-Path` is such a mess that I'm going to switch to `[io.path]::combine` which should be more consistent (so baring any other inconsistencies we should not have to worry about v5 or v6). And also, your latest screenshot looks like a bug somewhere in the concatenation. Tracking this down.

ctstone:
@gabog, new PR at #897. It would be good to get your feedback on this fix. Still PS 6 only -- once the new PR is cleared I will address PS 5 compatibility)

gabog:
sure thing, let me try it out and will post back on the PR.

PR #1004, opened 2019-03-20
SkillAdapter, SkillDialog and tests for remote skill invocation

darrenj:
# Remote Invocation

- Added a stripped down SkillDialog class that has the basic skill invocation for remote skills and all InProc activation removed. Authentication support is deliberately omitted as we refine the pattern for this.
- skillBegin event is sent to the Skill as today and deliberately does not support parameters as we refine the Orchestration "slots" pattern
- New SkillAdapter which can be attached to a new MVC controller for a Skill enabling remote invocation via the SkillDialog
- SkillAdapter doesn't send typing activities back when channelid is set to test which avoids random typing indicators breaking the test.
- Unit Tests that work against a (manually) adjusted POI skill already deployed to validate initial flows
- Simplified SkillDefinition class as a placeholder before manifest support, this has an Endpoint parameter instead of Assembly and doesn't have parameter/skillconfiguration
- SkillController that provides the HTTP->Activity and Activity->HTTP logic, this is configured on a Skill along with the SkillAdapter to ensure a Skill can be addressed via HTTP.

# Next Steps

- Implement Channel Token validation in the SkillAdapter
- ServiceUrl "post back" from the Adapter to the client asychronously
- Update Skills to leverage the new SkillControl/Adapter and deploy


lauren-mills:
I forgot to enforce the PR rule for admins. So congrats, the changes have been accepted 😄

PR #1085, opened 2019-04-10
[Botskills][TypeScript] Implement Skill CLI Tool

dfavretto:
## Description
- Implement new CLI Tool for manipulating Skills
    - Implement `list` command for listing all skills connected to the assistant
    - Implement `connect` command for adding a Skill to the assistant
    - Implement `disconnect` command for removing a Skill from the assitant

## Testing Steps
1. Go to `AI\lib\typescript\botskills\`.
1. Open a terminal in that location.
1. Run the command `npm install` to install dependencies.
1. Run the command `npm run build` to build the project.
1. Run the command `npm link` to symlink the package folder.
1. Use the Tool with any of the commands available, following the documentation present in `AI\lib\typescript\botskills\docs` on what options are available for every command.

## Checklist


lib/typescript/botskills/README.md
```diff
@@ -0,0 +1,40 @@
...
+- [Connect](./docs/connect-disconnect.md) a Skill to your assistant
+- [Disconnect](./docs/connect-disconnect.md) a Skill from your assistant
+- [List](./docs/list.md) all Skills connected to your assistant
+
+## Nightly builds
+Nightly builds are based on the latest development code which means they may or may not be stable and probably won't be documented. These builds are better suited for more experienced users and developers although everyone is welcome to give them a shot and provide feedback.
+
+You can get the latest nightly build of Botskills from the [BotBuilder MyGet]() feed. To install the nightly
+```bash
+npm config set registry https://botbuilder.myget.org/F/aitemplates/npm/
```
dfavretto, line 28: Maybe this link should be changed later

lib/typescript/botskills/src/botskills-connect.ts
```diff
@@ -0,0 +1,125 @@
...
+        return '';
+    });
+    process.exit(1);
+}
+
+function getManifest(): ISkillManifest {
+    // Determine wether the manifest will be taken locally or remotely
+
+    // Remote manifest
+    // PENDING
```
dfavretto: Here we'll be adding the logic to take a remote manifest file for the assistant

lib/typescript/botskills/src/botskills-connect.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+if (!skillManifest.endpoint) {
+    logger.error(`Missing property 'endpoint' of the manifest`);
+}
+if (!skillManifest.authenticationConnections) {
+    logger.error(`Missing property 'authenticationConnections' of the manifest`);
+}
+if (!skillManifest.actions || !skillManifest.actions[0]) {
+    logger.error(`Missing property 'actions' of the manifest`);
+}
```
dfavretto: We are validating some properties from the SkillManifest, but maybe @darrenj you can lead us to what properties should be validated as mandatory, or that need some special validation

dfavretto (commented):
Adding some comments on the code.

lib/typescript/botskills/src/botskills-connect.ts
```diff
@@ -0,0 +1,146 @@
...
+    });
+    process.exit(1);
+}
+
+async function getRemoteManifest(resourcePath: string): Promise<ISkillManifest> {
+    return get({
+        uri: <string> resourcePath,
+        json: true
+    });
+}
```
dfavretto: Now the remote Skill Manifest invocation is ready to be used

lib/typescript/botskills/src/botskills-connect.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (!skillManifest.endpoint) {
+        logger.error(`Missing property 'endpoint' of the manifest`);
+    }
+    if (!skillManifest.authenticationConnections) {
+        logger.error(`Missing property 'authenticationConnections' of the manifest`);
+    }
+    if (!skillManifest.actions || !skillManifest.actions[0]) {
+        logger.error(`Missing property 'actions' of the manifest`);
+    }
+}
```
dfavretto: The Manifest Schema validations were extracted to a function, but still have to be reviewed

lib/typescript/botskills/src/botskills-connect.ts
```diff
@@ -0,0 +1,146 @@
...
+    .action((cmd: program.Command, actions: program.Command) => undefined);
+
+const args: program.Command = program.parse(process.argv);
+
+if (process.argv.length < 3) {
+    program.help();
+    process.exit(0);
+}
+
+async function connectSkill(): Promise<void> {
```
dfavretto: Now all the `botskills connect` functionality is wrapped in an `async` function in order to use the remote skill manifest invocation with the requests, since it uses promises and need to be awaited

dfavretto (commented):
The comments on the implementation were updated

lib/typescript/botskills/README.md
darrenj: Yep - we can update with the real locations closer to build.

lib/typescript/botskills/src/botskills-connect.ts
darrenj: I sent a mail with manifest spec information. If a Skill doesn't have any actions is "empty" so that's a valid error. It's OK if a Action doesn't have any slots. I would stick to ensuring it has a name, endpoint, actions and actions have a source of utterances in some way (otherwise there's no way to trigger it.

lib/typescript/botskills/docs/connect-disconnect.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+### Options
+
+| Option                        | Description                                                                                                                                                 |
+|-------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| -l, --localManifest \<path>   | Path to local Skill Manifest file                                                                                                                           |
+| -r, --remoteManifest \<url>   | URL to remote Skill Manifest                                                                                                                                |
+| -a, --assistantSkills \<path> | Path to the assistant Skills configuration file                                                                                                             |
+| -d, --dispatchName \<name>    | Name of your assistant's '.dispatch' file                                                                                                                   |
```
darrenj: We should be able to make this optional (latest Powershell does this) as we can work this out from cognitivemodels file. Latest powershell accepts cognivemodels parameter (and defaults to cognitivemodels.json)

lib/typescript/botskills/docs/connect-disconnect.md
```diff
@@ -0,0 +1,62 @@
...
+botskills connect [options]
+```
+
+### Options
+
+| Option                        | Description                                                                                                                                                 |
+|-------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| -l, --localManifest \<path>   | Path to local Skill Manifest file                                                                                                                           |
+| -r, --remoteManifest \<url>   | URL to remote Skill Manifest                                                                                                                                |
+| -a, --assistantSkills \<path> | Path to the assistant Skills configuration file                                                                                                             |
```
darrenj: we can make this optional too - default to skills.json

lib/typescript/botskills/src/botskills-disconnect.ts
```diff
@@ -0,0 +1,89 @@
...
+);
+
+if (!skillToRemove) {
+    logger.warning(`The skill '${args.skillName}' is not present in the assistant Skills configuration file.
+Run 'botskills list --assistantSkills "<YOUR-ASSISTANT-SKILLS-PATH>"' in order to list all the skills connected to your assistant`);
+    process.exit(1);
+} else {
+    // Removing the skill manifest from the assistant skills array
+    logger.warning(`Removing the '${args.skillName}' skill from your assistant's skills configuration file.`);
+    assistantSkills.splice(assistantSkills.indexOf(skillToRemove), 1);
```
darrenj: we should also call dispatch remove otherwise we'll leave things in the dispatch configuration

darrenj (approved):
I've provided a few comments (likely just things in flight) so will approve and we'll discuss

lib/typescript/botskills/docs/connect-disconnect.md
dfavretto: The documentation was updated just after this, the dispatchName was made optional and defaults to the name located inside the `cognitivemodels` file, just like the Powershell script does

lib/typescript/botskills/docs/connect-disconnect.md
dfavretto: The documentation was updated just after this, the `assistantSkills` property was changed both in connect and disconnect commands for parity with the Powershell scripts, now its name is `skillsFile` and defaults to the `skills.json`

lib/typescript/botskills/src/botskills-disconnect.ts
dfavretto: This is what we plan to do for the disconnect as soon as we finish testing the connect functionality; mostly we want to check with a working assistant that the skill is connected **and** is being invoked successfully. After that we'll try disconnecting the skill, and it **shouldn't** throw an error for trying to reach the skill that we just disconnected

lib/typescript/botskills/src/utils/authenticationUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+
+            // Remove duplicate scopes
+            scopes = [...new Set(scopes)];
+            const scopeManifest: IScopeManifest[] = createScopeManifest(scopes);
+
+            // Update MSA scopes
+            logger.message('Configuring MSA app scopes ...');
+            let azureAppUpdateCmd: string = `az ad app update `;
+            azureAppUpdateCmd += `--id ${appSettings.microsoftAppId} `;
+            azureAppUpdateCmd += `--reply-urls "https://token.botframework.com/.auth/web/redirect" `;
```
dfavretto: Hey @lauren-mills, 
**FYI**: we had to add this parameter to the `az ad app update` command in order to ensure that we have a Reply URL for the Authentication configuration, we think this should be replicated in the Powershell Scripts.

lauren-mills: This should already be in the deploy.ps1 powershell script [here](https://github.com/Microsoft/AI/blob/199b99b1ef507ab6b13e2ea833d8460fddd81877/templates/Virtual-Assistant-Template/csharp/Template/VA/Deployment/Scripts/deploy.ps1#L77)

Let me know if I need to add it anywhere else

dfavretto: Oh sorry, didn't see that; however, that is being executed **only** when the `appId` is missing, so if the user sends an `appId` but didn't configure the `Reply URL`, we need to add it ourselves somewhere else.

I just pushed [some changes](https://github.com/Microsoft/AI/pull/1085/commits/67e0c514888dc0fb01b77d8d52cc4d1dead204e8) for getting the Reply URLs already present (if there are any) and to append this URL if it's not already present. 

Maybe this logic can be replicated in the `add_remote_skill.ps1` script just before the `az ad app update` command execution, what do you think?

PR #1098, opened 2019-04-12
Skill support for Slot Filling and Skill/Action wiring for SkillDialog

darrenj:
## Description

This PR might seem pretty straight forward but it does introduce a subtle but important change to how we wire up Skills registered in manifests.

- A Skill can have one or more actions, these actions have specific slots associated with them (parameters) and associated triggering utterances.
- Previously we grouped everything together at the dispatcher level under one skill but this means a Bot when identifying a skill for a given utterance cannot work out the Slots as it doesn't know the action.
- A change is required to our "AddSkill" CLI/powershell logic to not add one dispatch entry for the entire skill but instead one per action e.g. calendarSkill/AddMeeting, calendarSkill/summary.

We still add a SkillDialog per registerd Skill and rely on the Bot (VirtualAssistant template created) to pass the Action (the Dispatcher intent) via DialogOptions when calling BeginDialog. This ensures that a SkillDialog knows **both** the skill **and** Action. SkillDialog can then inspect slots, SkillContext and as appropriate pass across Slot data via the beginEvent.

I've introduced a new `SkillContext` class which acts as a propertybag of sorts for data. If a Bot puts data in here (e.g. username, location, etc.) then it's deliberately opting in to offer this data up for exchange with Skills). This SkillContext object is then used on the Skill side to hold the data and is created by the SkillMiddleware. Skills will need to take this data and place in it's own state representation for usage.

I've also made changes to SkillDialog to enable mocking and a set of fairly clever tests that enable us to validate the SkillDialog firstly works (sending a message to a faux endpoint) and whether the skilLBeginEvent has the right slot data based on the SkillManifest and what data is in SkillContext.

SkillMiddleware needs some further testing and tests written to validate this side, we can then add to each of our Skills and also our Skill template.

Feedback welcome!


lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillContext.cs
```diff
@@ -0,0 +1,46 @@
...
+
+            set
+            {
+                _contextStorage[name] = value;
+            }
+        }
+
+        public bool TryGetValue(string key, out object value)
+        {
+            return _contextStorage.TryGetValue(key, out value);
```
lzc850612, line 43: i changed the skills project to netstandard which means you won't be able to use TryGetValue function on a dictionary any more :( i know this is annoying but it's important that we publish the package as a netstandard instead of netcore

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
```diff
@@ -24,9 +25,10 @@ namespace Microsoft.Bot.Builder.Skills
     /// </summary>
     public class SkillDialog : ComponentDialog
     {
-        private static readonly HttpClient _httpClient = new HttpClient();
+        protected HttpClient _httpClient = new HttpClient();
         private MicrosoftAppCredentialsEx _microsoftAppCredentialsEx;
         private IBotTelemetryClient _telemetryClient;
+        private UserState _userState;
         private JsonSerializerSettings _serializationSettings = new JsonSerializerSettings
```
lzc850612: this can go away because there's a Serialization class in the solutions lib and it's got the static variable that we can use.

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillMiddleware.cs
```diff
@@ -10,17 +11,25 @@ namespace Microsoft.Bot.Builder.Skills
     /// </summary>
     public class SkillMiddleware : IMiddleware
```
lzc850612, line 12: the SkillMiddleware, yes!!!

templates/Virtual-Assistant-Template/VirtualAssistantTemplate/VirtualAssistantTemplate/Dialogs/MainDialog.cs
```diff
@@ -54,9 +54,15 @@ public class MainDialog : RouterDialog
...
 
+            // Each Skill has a number of actions, these actions are added as their own SkillDialog enabling
+            // the SkillDialog to know which action is invoked and identify the slots as appropriate.
             foreach (var skill in settings.Skills)
             {
-                AddDialog(new SkillDialog(skill, _responseManager, new MicrosoftAppCredentialsEx(_microsoftAppCredentials.MicrosoftAppId, _microsoftAppCredentials.MicrosoftAppPassword, skill.MSAappId), telemetryClient));
+                // Each action within a Skill is registered on it's own as a child of the overall Skill
+                foreach (var action in skill.Actions)
+                {
+                    AddDialog(new SkillDialog(skill, action, _responseManager, new MicrosoftAppCredentialsEx(_microsoftAppCredentials.MicrosoftAppId, _microsoftAppCredentials.MicrosoftAppPassword, skill.MSAappId), telemetryClient, userState));
```
lzc850612: couple of things about this approach:
1. the number of SkillDialog will multiply pretty quickly as we add more skills into a bot
2. i still feel we need one SkillDialog per skill, and maybe manage different actions inside it. right now it seems fine but logically SkillDialog should be the one that's managing everything inside a skill including Actions (it's probably fine if we want to manage them as dialogs). it's weird to have EmailSkill/SendEmail and CalendarSkill/BookAMeeting skillDialogs on the same level. 
3. also the VA's maindialog should not do stuff about the skill's actions. maybe we should push this piece of initialization code down to the Skills lib which is a better pattern for VA developers.

lzc850612 (changes requested):

templates/Virtual-Assistant-Template/VirtualAssistantTemplate/VirtualAssistantTemplate/Dialogs/MainDialog.cs
darrenj: Actually, on reflection your right. The Dispatcher has to be aware of the different actions but one central skilldialog per skill will work just fine. Let me refactor - great point

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillContext.cs
darrenj: Seems OK :)

templates/Virtual-Assistant-Template/VirtualAssistantTemplate/VirtualAssistantTemplate/Dialogs/MainDialog.cs
darrenj: I've moved to using the DialogOptions on BeginDialog to pass the ActionId, the calling Bot knows the "actionid" as it will == dispatcher Intent.

darrenj:
Updated PR text to reflect the tweak

lzc850612 (approved):

PR #1119, opened 2019-04-18
Implement dependency injection for Dialogs in Enterprise Bot Template

hansmbakker:
## Description
There has been a lot of talk and wishes for improvement around dependency injection of Dialogs.
- https://github.com/Microsoft/botbuilder-dotnet/issues/860
- https://github.com/Microsoft/botbuilder-dotnet/issues/357
- https://github.com/Microsoft/botbuilder-dotnet/issues/1188
- https://stackoverflow.com/questions/52549850/how-do-you-do-dependency-injection-in-a-class-that-inherits-dialogcontainer-bot

I believe Enterprise Bot Template should set a good example. I'm trying to do that with this PR without changes to the bot framework.

### Architecture
This PR builds the existing ASP.Net Core DI code and follows the approach from https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/42.scaleout/Startup.cs

- Dialogs are registered in a separate method `RegisterDialogs` in `Startup.cs`
- The `MainDialog` is injected in `Bot`
- Redundant constructor parameters are removed:
  - Some dialogs do not need `UserState` or `ConversationState` anymore
  - UserState and ConversationState can be injected directly because they are registered
  - `OnboardingDialog` generate its own state accessor based on the injected `UserState`

## Testing Steps
Run the bot as before

## Checklist

### DIFF
--- a/templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/EnterpriseTemplate/Bot.cs
+++ b/templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/EnterpriseTemplate/Bot.cs
@@ -18,8 +18,6 @@ namespace $safeprojectname$
     {
         private readonly BotServices _services;
         private readonly ConversationState _conversationState;
-        private readonly UserState _userState;
-        private readonly IBotTelemetryClient _telemetryClient;
         private DialogSet _dialogs;
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


darrenj:
Thanks! We totally agree and actually have this almost ready for our Build release, would love your feedback on whether it hits the Mark - it's a big step forward we feel. It's all in the 4.4 branch and we should have a new vsix today/tomorrow if you'd like to test?

At the same time you'll see the new skill work!

hansmbakker:
@darrenj
That's really nice to hear, and I'm curious to see the results of that!

At the same time I would really appreciate it then, if you could maintain a kind of roadmap with a clear indication of what is being worked on and where that can be found.

That would allow people to collaborate if possible, and it would prevent people from spending effort and time on PRs that are redundant...

E.g. [@johnataylor briefly mentioned that there would be worked on dependency injection](https://github.com/Microsoft/botbuilder-dotnet/issues/1188#issuecomment-458689974) but did not give any details and did not reply to questions anymore. This made it appear to be a longer-term thing that might even not be released, so I came up with this PR based on the mentioned sample.

However if I had known that you are concretely working on an implementation that is more thorough, maybe I would not have made this PR.

lauren-mills:
Closing since this logic is being handled in the 4.4 branch.

hansmbakker:
@lauren-mills @darrenj if I look now in the 4.4 branch for the Enterprise Bot Template I do not see any dependency injection of dialogs in the `MainDialog` for example. The `new` keyword is still used for e.g. the `OnboardingDialog`.
https://github.com/Microsoft/AI/blob/4.4/templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/EnterpriseTemplate/Dialogs/Main/MainDialog.cs

Am I looking in the wrong place?

darrenj:
Virtual Assistant template replaces the Enterprise Template: https://github.com/Microsoft/AI/tree/4.4/templates/Virtual-Assistant-Template

hansmbakker:
OK, thank you for the clarification!

PR #1122, opened 2019-04-19
Fixed bug with Locale Middleware

tiagodenoronha:
## Description

Even if you didn't define the locale middleware, the condition would always be true since the context.Activity.Locale returned string.empty and not null.
This PR fixes the issue.

### Bug Fixes
Locale not getting caught correctly. Should I open an issue before?

## Testing Steps
Define a defaultLocale in appsettings and define some utterances for that locale and run the bot.

## Checklist


lauren-mills:
Thanks for your PR! We are moving this middleware into a nuget package and have already made this change. You can see the latest here: https://github.com/Microsoft/AI/blob/master/lib/csharp/microsoft.bot.builder.solutions/microsoft.bot.builder.solutions/Middleware/SetLocaleMiddleware.cs

Stay tuned for a new template coming soon!

tiagodenoronha:
Oh, got it! Thanks for the heads up! :)

PR #1195, opened 2019-04-30
Update training utterances for automotive skill

darrenj:
- Updated LUIS models with some additional examples found to be weak with combined dispatch (qna/chit-chat)
- Removed profanity

@lauren-mills I have made this in the AI branch as that's where the deployment files are, assume when you do a merge for the new skill deployment stuff you'll pick this up.


lzc850612 (approved):
...

PR #1249, opened 2019-05-03
[build] Use different versions for CLI tool and botbuilder-* libraries

enzocano:
## Description
Add new params to use separate versions for the `CLI tool` and `botbuilder-*` packages. The `template` uses the same version of the `botbuilder-*` because it depends on them.

New params
`-cliversion` for the CLI tool
`-pkgversion` for the other projects

1. Open powershell in the root directory of the project.
1. Run `./build.ts.ps1 -cliversion 4.4.0 -pkgversion 1.0.0`
1. You will see the packages in `./outputpackages`

### DIFF
--- a/build.ts.ps1
+++ b/build.ts.ps1
@@ -1,9 +1,15 @@
 ﻿Param(
-	[string] $version
+	[string] $pkgversion,
+	[string] $cliversion
 )
 
-if (-not $version) {
-    Write-Host "Version required!.  Please use the param -version" -ForegroundColor DarkRed
+if (-not $pkgversion) {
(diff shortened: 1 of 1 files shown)


build.ts.ps1
```diff
@@ -36,7 +42,7 @@ popd
 
 pushd .\templates\Virtual-Assistant-Template\typescript\generator-botbuilder-assistant
 
-npm version $($version) --allow-same-version
+npm version $($pkgversion) --allow-same-version
```
lzc850612: this should be $(cliversion) as well

lzc850612 (changes requested):

build.ts.ps1
enzocano: fixed!

lzc850612 (approved):

PR #1279, opened 2019-05-05
fix for LU files that failed stricter ludown parsing in 1.3.0

darrenj:
LUdown 1.3.0 along with a bug around patterns had stricter checks which tripped up on pointofinterest.lu


darrenj:
@vishwacsena FYI and thanks for the spot!

lzc850612 (approved):

PR #1352, opened 2019-05-13
set recipient id to be the skill id when calling skill

lzc850612:
## Description

When calling skill, the recipient id should be the skill's id.

## Checklist


ryanisgrig (approved):

darrenj:
How does a Skill make use of this?

lzc850612:
> How does a Skill make use of this?

AFAIK we don't do anything now. It's part of the protocol so when a bot calls a skill, it should be the skill's id instead of the bot. Right now channel does this so we need to do this in the Skills pkg when relaying messages to skills. I don't think it's one of the IDs we need to munge as per our discussion yesterday.

PR #1376, opened 2019-05-17
Yuesu/add handling for skillcontext

KayMKM:
### Bug Fixes
remove the Slot constructor to avoid can't load slot from manifest.
add type convert logic when skill get timezone from activity.

## Checklist

### DIFF
--- a/lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/Models/Manifest/Slot.cs
+++ b/lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/Models/Manifest/Slot.cs
@@ -8,11 +8,11 @@ namespace Microsoft.Bot.Builder.Skills.Models.Manifest
     /// </summary>
     public class Slot
     {
-        public Slot(string name, List<string> types)
-        {
-            Name = name;
-            Types = types;
-        }
+        //public Slot(string name, string[] types)
+        //{
+        //    Name = name;
+        //    Types = types;
+        //}
 
         [JsonProperty(PropertyName = "name")]
         public string Name { get; set; }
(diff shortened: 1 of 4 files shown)


lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/Models/Manifest/Slot.cs
```diff
@@ -8,12 +8,6 @@ namespace Microsoft.Bot.Builder.Skills.Models.Manifest
     /// </summary>
     public class Slot
     {
```
darrenj, line 10: This should not be removed, used elsewhere in the overall solution?

darrenj (changes requested):
Events will not be sent by VA to the skill. skillcontext instead passes this across and the skillmiddleware makes this available to the skill. Each skill needs to implement populatestatefromskillcontext as per https://github.com/microsoft/botframework-solutions/blob/master/skills/src/csharp/pointofinterestskill/pointofinterestskill/Dialogs/MainDialog.cs

Here you will get timezone if passed

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/Models/Manifest/Slot.cs
darrenj: You can remove this change, @lzc850612 has made some fixes 😎

darrenj:
> Events will not be sent by VA to the skill. skillcontext instead passes this across and the skillmiddleware makes this available to the skill. Each skill needs to implement populatestatefromskillcontext as per https://github.com/microsoft/botframework-solutions/blob/master/skills/src/csharp/pointofinterestskill/pointofinterestskill/Dialogs/MainDialog.cs
> 
> Here you will get timezone if passed

Actually if this is for local mode testing that's fine!

PR #1389, opened 2019-05-21
fix skill middleware handling of SkillContext

lzc850612:
## Description

In our SkillMiddleware, it cannot convert the activity.Value to SkillContext directly. We need to explicitly deserialize it. 

It appears the Slot definition's constructor will cause the Slots not being deserialized properly. Removing it to fix the issue.

## Checklist


darrenj:
Can we just not add a default constructor for serialization purposes? Otherwise I think you'll break other parts of the implementation?

darrenj (changes requested):
Can we just not add a default constructor for serialization purposes? Otherwise I think you'll break other parts of the implementation?

lzc850612:
> constructor

removing the constructor will break more. There's a lot of code using that constructor.

why do you think it will break other parts? I would think explicitly deserializing here is a clearer thing to do rather that casting from a string that's deserialized over the wire which is not gonna work.

darrenj:
The constructor with parameters was deliberately put there to simplify the slot implementation if I recall correctly, if you need a default constructor for serialization then add that in addition rather than removing?

darrenj (approved):

PR #1468, opened 2019-06-03
[Speech] Remove dependency on ConnectorClient for MultiProviderAuth

darrenj:
Remove dependency on `ConnectorClient` which isn't present unless you use `BotFrameworkAdapter`, if MicrosoftAppSettings is passed on the constructor to `MultiProviderAuthChannel` and the Speech Channel is being used then the credentials are used on the call to the OAuthClient to retrieve tokens.

The previous approach used ConnectorClient to get this which is now removed.

This line in Startup.cs of your VA:
```
return new MultiProviderAuthDialog(oauthConnections);
```

Need to be replaced with the following in scenarios where Speech is being used. We can explore making this part of the base template and not manual moving forward.

```
return new MultiProviderAuthDialog(oauthConnections, new MicrosoftAppCredentials(settings.MicrosoftAppId, settings.MicrosoftAppPassword));
```


ryanisgrig (approved):
When it's officially out of preview, we should consider the whole set of Streaming Extensions changes to be in the default template.

PR #1480, opened 2019-06-04
[Botskills] Implement refresh command

Batta32:
## Description

- Add `refresh` command.
- Use of the new command in `connect` & `disconnect` implementation.
- Add tests for the `refresh` command.
- Update `disconnectTest`.
- Add/Update documentation related to `refresh` command.
- Previously named `train` command renamed to `refresh`.

## Testing Steps
**Using botskills**

1. Connect a Skill to the Assistant follow this [steps](https://github.com/microsoft/botframework-solutions/blob/fae9975165e820c00761d23782b8622abfef6313/lib/typescript/botskills/docs/connect-disconnect.md) **with the `--noRefresh` flag**
1. Run `botskills refresh--<cs | ts> ` (depends on your assistant language)
1. Check that the refresh command updates your **dispatch model**
1. Go to `AI\lib\typescript\botskills`
1. Run `npm run test`
1. Check that all tests are passing successfully

image

## Checklist
- [X] I have added or updated the appropriate unit tests
- [X] I have updated related documentation


darrenj (commented):
I'm struggling a little bit with the name train, perhaps dispatchRefresh is better? @lauren-mills thoughts?

Batta32:
Hi @darrenj!
There is no problem to rename the command, but besides the behavior of updating the _dispatch model_, this includes the execution of _luisgen_, and we want to keep the command names as simpler as possible, **making them 1 word long**.
Taking your input into account, we considered naming it just _Refresh_, and the _Refresh_ command that we are working on (which we'll take in a manifest already connected, disconnect it and reconnect again) could be renamed to Update. With this in mind, **we would have the following 2 commands**:
	• **Refresh** - Used to take the local Dispatch model (which was generated in the **connect/disconnect** command if it was executed with the noRefresh argument) and execute **dispatch refresh** with that model, and then **luisgen** command for updating the _DispatchLuis_ model for the VA.
	• **Update** - Used to update the connection to a specific Skill. This is useful when the Skill is already connected and suffered some changes, and needs to be disconnected and then connected again, finally, the model is trained again with the aforementioned command **Refresh**.

Thoughts?

dfavretto:
Hey @darrenj, we updated this PR with all necessary changes, and now is ready for review and merge! 😃

darrenj (approved):

PR #1501, opened 2019-06-07
Fixes path issue with dispatch sync

BeigeBadger:
## Description
Fixes an issue introduced by https://github.com/microsoft/botframework-solutions/commit/6061da900dcf1bf60caacd487a763cadf50fa4bd where the dispatch file would be updated in the root of the dispatch directory, but the dispatch file used for dispatch refresh is located under dispatch/locale, meaning that running the `update_cognitive_models.ps1` script didn't work.

### Bug Fixes
As it stands the current script would create/update a .dispatch file in `Deployment\Resources\Dispatch` rather than `Deployment\Resources\Dispatch\locale-folder` which means that when the dispatch refresh command is run (and looks for a dispatch file inside the locale-folder) it won't be the file that was just updated. This change updates the behaviour so that we are both writing to, and reading from the locale-folder.

## Testing Steps
Tested against a KB and LUIS app that I created


BeigeBadger:
@NathanNZ here is my PR that I made to fix up the dispatch stuff.

lauren-mills (approved):

Issue #1528, opened 2019-06-12
[Linked Accounts] Enable companion app scenario


ryanisgrig:
**Is your feature request related to a problem? Please describe.**

Linked Accounts web app should show an example of how to use in an Android companion app.

**Describe the solution you'd like**

Provide 

Home page redirects to Linked Accounts page with optional boolean query parameter "companionApp." If true, it will show a new "Complete and return to app" button that opens the deep link "microsoft://virtualassistant/user/{userId}.

PR #1539, opened 2019-06-12
RequestOAuthFlow event in Direct Line Speech auth scenario

ryanisgrig:
### Maintenance
If a token is missing for a user using the direct line speech channel, their client will receive an event named "RequestOAuthFlow"
It's up to the client to configure this with their sign-in scenario

#1528 

## Testing Steps

On the DL Speech client, trigger a productivity skill and verify that the VA sends both a message telling them to log in as well as an event named RequestOAuthFlow

## Checklist


ryanisgrig:
I'm open to this being configurable, but essentially this will allow for a user to use a VA Android client and:
1. Trigger skill requiring auth
2. VA cannot find a stored token, tells user to log in
3. Client receives RequestOAuthFlow event and opens Linked Accounts web app in the view (with companionApp=true query parameter)
4. User authenticates with web app and then the VA
5. User clicks "Return and complete in app", which contains the deep link "microsoft://virtualassistant/user/{userId}"
6. Client is set up to open with this URL and stores the user id in it's settings

darrenj (approved):

PR #1540, opened 2019-06-13
[Skills] fix the issue where in webchat skill responses are not sent back to user properly

lzc850612:
## Description

Right now for Webchat, there's a weird issue where calls to skill won't be sent back to user. It's because we tempered with the Recipient id of the activity before sending it to the skill to set it to be the skill's msa id. It has to remain as the current bot's id for the activity sending to work. So set it back to original value after call to skill is constructed to fix the issue.

also added logger to log the exception on SkillCallingRequestHandler class since we can't throw the exception back because it's an API call.

## Checklist

### DIFF
--- a/lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillWebSocketTransport.cs
+++ b/lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillWebSocketTransport.cs
@@ -20,20 +20,20 @@ namespace Microsoft.Bot.Builder.Skills
         private IStreamingTransportClient _streamingTransportClient;
         private readonly SkillManifest _skillManifest;
         private readonly IServiceClientCredentials _serviceClientCredentials;
-		private readonly IBotTelemetryClient _botTelemetryClient;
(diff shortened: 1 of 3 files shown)


lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillWebSocketTransport.cs
```diff
@@ -44,29 +47,30 @@ public async Task<bool> ForwardToSkillAsync(ITurnContext turnContext, Activity a
...
+            // set recipient to the skill
+            var recipientId = activity.Recipient.Id;
+            activity.Recipient.Id = _skillManifest.MSAappId;
 
             // Serialize the activity and POST to the Skill endpoint
             var body = new StringContent(JsonConvert.SerializeObject(activity, SerializationSettings.BotSchemaSerializationSettings), Encoding.UTF8, SerializationSettings.ApplicationJson);
             var request = Request.CreatePost(string.Empty, body);
+
+            // set back recipient id to make things consistent
+            activity.Recipient.Id = recipientId;
```
lzc850612, line 72: this is the main fix

darrenj (approved):

PR #1571, opened 2019-06-17
Fix project folder parameter in bot deploy script

### LINKED ISSUE #1570: Unable to call deploy script from outside project root
## Project

**Name**: Virtual Assistant

**Language**: C# 

## Description
If launching `deploy.ps1` from context outside of root of the project, the `projDir` property does not get passed down to `publish.ps1` where deploying the web bot will fail.

## To Reproduce
From a parent directory, script a call to `deploy.ps1` including the `-projDir` flag. An error will be displayed in PowerShell at the `az bot prepare-deploy` step.

## Expected behavior
The `$projDir` value gets passed from `deploy.ps1` to `publish.ps1` as needed. (It does get passed to `deploy_cognitive_models.ps1`, for example.)

mitch-b:
## Description

I had built a parent PowerShell file to deploy a C# Virtual Assistant bot abstracted from the script bundled with the template. However, the `projDir`/`projFolder` parameter is not passed through from `deploy.ps1` to `publish.ps1` - including usage of a misspelled variable.

Closes #1570

### Bug Fixes

When deploying the Virtual Assistant C# template, `deploy.ps1` will now pass along the `$projDir` parameter into `publish.ps1`. Additionally, a variable name was fixed from `$prjFolder` to `$projFolder` which produced erroneous behavior.

### DIFF
--- a/templates/Skill-Template/csharp/Sample/SkillSample/Deployment/Scripts/deploy.ps1
+++ b/templates/Skill-Template/csharp/Sample/SkillSample/Deployment/Scripts/deploy.ps1
@@ -197,7 +197,7 @@ if ($outputs)
 	Invoke-Expression "$(Join-Path $PSScriptRoot 'deploy_cognitive_models.ps1') -name $($name) -luisAuthoringRegion $($luisAuthoringRegion) -luisAuthoringKey $($luisAuthoringKey) -outFolder `"$($projDir)`" -languages `"$($languages)`""
 	
 	# Publish bot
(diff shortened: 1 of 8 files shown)


darrenj:
Thanks for reporting and resolving! Could you also apply this change to the sample project in https://github.com/microsoft/botframework-solutions/tree/master/templates/Virtual-Assistant-Template/csharp/Sample

Same with the skill template / sample too if possible. We are working on a build process that pulls these from a central location.

darrenj (approved):

darrenj:
Thanks Barry!

PR #1577, opened 2019-06-17
Update known issue around Teams and VA change for existing projects using skills

darrenj:
- SDK 4.4.5 and latest Solutions package addresses Teams issues so updating docs
- Adding known issue about line to remove from VA to avoid Skill invocation message being sent twice.

### DIFF
--- a/docs/reference/knownissues.md
+++ b/docs/reference/knownissues.md
@@ -2,46 +2,9 @@
 
 ## The Teams channel doesn't render OAuth cards.
 
-There is a known issue in the Teams channel where the default OAuth Card is not shown when using the OAuthPrompt in the SDK. To workaround this in the short-term please follow the following instructions.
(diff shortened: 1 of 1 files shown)


docs/reference/knownissues.md
```diff
@@ -2,59 +2,27 @@
...
-
-        return result;
-    }
-    else
-    {
-        return await base.OnContinueDialogAsync(innerDc, cancellationToken);
-    }
-}
+  // Pass the activity we have
+  var result = await dc.ContinueDialogAsync();
```
lzc850612: i think it's better to say

Remove the calling of 'ContinueDialogAsync' by changing the existing code:

await dc.BeginDialogAsync(identifiedSkill.Id);

// Pass the activity we have
var result = await dc.ContinueDialogAsync();

To

var result = await dc.BeginDialogAsync(identifiedSkill.Id);

lzc850612 (changes requested):

Issue #1508, opened 2019-06-07
Use ChoiceFactory when creating cards


EricDahlvang:
## User Story
As bot developer
I want to use ChoiceFactory to create choice prompts
so that the UI experience will automatically adjust for all the different channels

## Acceptance Criteria 
*List your requirements.*
- [ ] Consistent use in VA, Skill Template, and Skills
- [ ] ...
- [ ] ...

## Resources
*Provide any screenshots or conversation mockups to support your request.*
*Learn how to mock up bot conversations with [.chat files](http://aka.ms/bfchatfile).*

## Notes
*Provide any additional information.*

**Is your feature request related to a problem? Please describe.**
Some channels do not support SuggestedActions.  ChoiceFactory will attempt to determine if the channel supports them, and if not it will create something the channel is able to render.

**Describe the solution you'd like**
Instead of creating [SuggestedActions](https://github.com/microsoft/botframework-solutions/blob/master/templates/Enterprise-Template/src/csharp/EnterpriseBotTemplate/EnterpriseTemplate/Dialogs/Main/MainResponses.cs#L89) directly, use the [ChoiceFactory](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.Dialogs/Choices/ChoiceFactory.cs) which will create a card based on the channel.

xieofxie:
It is very useful. Is it possible to use the logic in ForChannel to check [ListStyle](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.Dialogs/Choices/ListStyle.cs)? So fallback could be used in functions like [AppendChoices](https://github.com/microsoft/botbuilder-dotnet/blob/3397372171a188821f1ef3d65d6e92bee3b30965/libraries/Microsoft.Bot.Builder.Dialogs/Prompts/Prompt.cs#L141).

darrenj:
Merged

Issue #1617, opened 2019-06-20
[POISkill] MS Teams: Exception thrown: 'Microsoft.Bot.Schema.ErrorResponseException'


samaea:
## Project

**Name**: Point of Interest Skill

**Language**: C#

## Description
When CompleteAsync() is called when using Microsoft Teams as the channel, the 'Microsoft.Bot.Schema.ErrorResponseException' exception is thrown due to a HTTP Bad Request when calling the endpoint https://smba.trafficmanager.net. I think the issue is to do with MS Teams not supporting the activity type "endOfConversation":- https://github.com/microsoft/botframework-sdk/issues/3300 and https://github.com/microsoft/botframework-sdk/issues/4774.

## To Reproduce
1. Use NGROK and set the bot service messaging endpoint to that URL.
2a) Type in "what is nearby" and complete the dialog
OR
2b) Type in "cancel".
3) You will receive an IM from the bot stating something along the lines of "Sorry, something went wrong!".

## Expected behavior
No error messages should be displayed and for the conversation state to be cleared.

## Screenshots
![Bot Schema ErrorResponseException](https://user-images.githubusercontent.com/32162601/59854959-de304580-936b-11e9-9085-5141986278f3.png)

## Additional context
- Using latest version of VS Studio 2019.
- Latest version of NGROK.
- On a brand new Windows 10 VM image.

#### Request
``` 
POST https://smba.trafficmanager.net/amer/v3/conversations/a%3A1kXm6XSjgHQha0bGz1xhd8AZqLYF0bjxc3TuqzwKGDkDboL_qzBtwIx_W9GOfdMGmUK-A3suqhn8aa7ggEiunGfEGzq6p81yzFVKCyDyp1328dthHoiRp-YLxDK6BoEzQ/activities/1561038373128 HTTP/1.1
User-Agent: Microsoft-BotFramework/3.1 BotBuilder/4.0.0.0 (.NETCoreApp,Version=v2.2; Microsoft Windows 10.0.17763 ; X64)
Content-Type: application/json; charset=utf-8
Content-Length: 888
Host: smba.trafficmanager.net

{
  "type": "endOfConversation",
  "timestamp": "2019-06-20T13:46:16.0955372+00:00",
  "serviceUrl": "https://smba.trafficmanager.net/amer/",
  "channelId": "msteams",
  "from": {
    "id": "28:64fbf7f0-1177-4b3d-a0cc-543c27ed9897",
[...]

PR #1646, opened 2019-06-24
Update skills to render in Teams (#1394)

### LINKED ISSUE #1394: Update default Adaptive Cards to render with a client-agnostic support
Testing on the Adaptive Cards Android SDK has revealed rendering issues that are not seen on the default Emulator. Updating with the following changes will allow cards to look more native to their platform.

Closes #1467 

* Remove the root-level backgroundImage (white background)
* Remove any instance of root-level `"height":"stretch"` (this is invalid)
* Replace all `Textblock.Color:dark` instances with `Textblock.Color:default`
* Resolve invalid base64 payloads (only supports characters ‘A-Z’, ‘a-z’, ‘0-9’, ‘+’ and ‘/’). Remove instances of  "%0D%0A" which are line breaks.
    * `data:image/svg+xml;` and `data:image/png+xml;` can remain
* Update the `Microsoft.Bot.Builder.Solutions` and `Microsoft.Bot.Builder.Skills` packages to use Adaptive Cards v1.2 (there will be a future update to take advantage of v1.2 properties)
* Use unique id names for each card property. For example, the Point of Interest cards have two objects with `"id":"name"`. This breaks validation on the new v1.2 renderer.
* Move the backgroundImage (blue header) in the first column to the first container.
**Before**
image
**After**
image

* Validate that the new cards render correctly in [App Studio for Microsoft Teams](https://docs.microsoft.com/en-us/microsoftteams/platform/get-started/get-started-app-studio)

_See Bing Search Skill PR I've written for example_

**Skills**
- [x] Calendar
- [x] To-Do
- [x] Email
- [x] Weather
- [x] News
- [x] Search
- [x] Point of Interest
- [x] Automotive

### LINKED ISSUE #1467: Teams not rendering Adaptive Cards used in Productivity and PoI Skills
## Project

Skills

## Description

When using the Productivity and PoI skills within Teams the Adaptive Cards do not render at all, the skills are still functional but lacking the visual cues that are key to certain tasks. Other channels, e.g WebChat work fine.

## To Reproduce

Use VA/Skills within Teams

## Expected behavior
- Cards to render, down-leveling if needed.

## Screenshots
image

## Additional context
- Working with the Teams team to ascertain the underlying issue and identifying a workaround (change the adaptive card JSONs to "down-grade" them.

xieofxie:
* update package version

* provide v1.0 json of adaptive cards for Teams

* skip certain responses that Teams could not handle

* separate certain prompts that Teams could not handle

* use svg in weather and todo skills for Teams

* update related tests

* update md and some htmls

## Purpose

Show skills in Teams correctly.

## Changes

*.json are changed according to #1394 

In Teams,

1. use *.1.0.json if necessary
    - All data images are removed. Images of weather and todo are kept as svg
    - Textblock.Color is changed to Default from Light if once above some images
2. ListStyle.SuggestedAction is not shown
3. AdaptiveCard is sent separately from Prompt since in ListStyle.Auto, prompt may be shown as HeroCard
4. Skip certain responses

## Tests

Covered by existing tests.

## Feature Plan

These changes will be revoked when the following are updated:

1. Teams supports adaptive card v1.2
2. Manually set ListStyle.SuggestedAction could fallback to ListStyle.Auto following logic in ChoiceFactory #1508 
3. [HeroCard could have AdaptiveCard](https://github.com/microsoft/botbuilder-dotnet/issues/2102)
4. Teams supports ActivityTypes.EndOfConversation, ActivityTypes.Event


xieofxie:
Do we have some method to quickly go through all the possible conversations of a skill/bot? Currently I manually input sentences to verify the cards..

darrenj:
@xieofxie - Thanks for doing this. Yes, there is an easier way to test.

Within Teams you can use the [App Studio](https://docs.microsoft.com/en-us/microsoftteams/platform/get-started/get-started-app-studio) and then the Card Editor to paste in JSON and crucially you must click "Send Card" - if it appears in a Chat it should work fine, if it doesn't appear when you send it doesn't!

skills/src/csharp/automotiveskill/automotiveskill/Dialogs/MainDialog.cs
```diff
@@ -114,10 +117,15 @@ protected override async Task RouteAsync(DialogContext dc, CancellationToken can
 
         protected override async Task CompleteAsync(DialogContext dc, DialogTurnResult result = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            var response = dc.Context.Activity.CreateReply();
-            response.Type = ActivityTypes.EndOfConversation;
+            // workaround. if connect skill directly to teams, the following response does not work.
+            if (dc.Context.Adapter is IRemoteUserTokenProvider remoteInvocationAdapter || Channel.GetChannelId(dc.Context) != Channels.Msteams)
```
lzc850612, line 121: why checking if the adapter is IRemoteUserTokenProvider?

skills/src/csharp/automotiveskill/automotiveskill/Dialogs/VehicleSettingsDialog.cs
```diff
@@ -518,6 +534,12 @@ private string UnitToString(string unit)
         /// <returns>A Task.</returns>
         private async Task SendActionToDevice(WaterfallStepContext sc, SettingChange change)
         {
+            // workaround. if connect skill directly to teams, the following reply does not work.
+            if (!(sc.Context.Adapter is IRemoteUserTokenProvider remoteInvocationAdapter) && Channel.GetChannelId(sc.Context) == Channels.Msteams)
```
lzc850612, line 538: why checking if adapter is IRemoteUserTokenProvider?

lzc850612 (changes requested):
Same comment for the extra checking of IRemoteUserTokenProvider in all files.

skills/src/csharp/automotiveskill/automotiveskill/Dialogs/MainDialog.cs
xieofxie: I want to skip sending these types of Activity directly to Teams because it will cause 'An error occurred. Give me some time and try again later.'. If it is IRemoteUserTokenProvider, then the skill is connected via Virtual Assistant and it will be fine.

darrenj: Yes - this is related to the #1617 raised last week whereby skills send EndOfConversation even if used "on their own" i.e. without VA. This causes Teams to throw errors so we need to omit sending these unless it's in VA skill mode.

xieofxie: So we should omit sending EndOfConversation  in *all* channels when without VA? I could patch that in this pr

skills/src/csharp/automotiveskill/automotiveskill/Dialogs/VehicleSettingsDialog.cs
xieofxie: Same as the reason above. But what is the expected behavior of ActivityTypes.Event? Thanks

skills/src/csharp/automotiveskill/automotiveskill/Dialogs/MainDialog.cs
darrenj: @lzc850612 - could this be better suppressed within the adapter infrastructure to do you think? It would simplify things if so?

lzc850612: just curious, can we ask Teams to handle EndOfConversation properly? is that possible?

darrenj: We can ask them yes but won't be a quick change.

darrenj (approved):

lzc850612 (approved):

Issue #929, opened 2019-03-05
Phone Skill (Expand Automotive Domain)


darrenj:
## User Story
As ...
I want ...
so that ...

## Acceptance Criteria 
*List your requirements.*
- [ ] ...
- [ ] ...
- [ ] ...

## Resources
*Provide any screenshots or conversation mockups to support your request.*
*Learn how to mock up bot conversations with [.chat files](http://aka.ms/bfchatfile).*

## Notes
*Provide any additional information.*

## Phone Skill

In pull request (#1659) and will provide a separate skill for calling scenarios including telephone number and lookup from Google/Email calendars as per the Email skill approach

outofthecave:
Shouldn't Phone, Media, and Tuner be separate skills?  I think they might also be useful outside of an automotive scenario.

Looks good to me otherwise.

darrenj:
Delivered

PR #1659, opened 2019-06-25
Phone skill

outofthecave:
Part of https://github.com/microsoft/botframework-solutions/issues/929

## Purpose

New Phone skill

## Changes

If the changes aren't easy to understand, please let me know, then I need to document them better :)

## Tests

Tests were added for various flows.

## Feature Plan

Here's what's left to do:
- [x] Get the deployment working out of the box.
  * <del>It currently has an issue with the Ludown tool (https://github.com/microsoft/botbuilder-tools/issues/1177), which has been fixed and merged into master (https://github.com/microsoft/botbuilder-tools/pull/1178), but not yet released as far as I know.</del>
  * <del>There might be authentication issues, but it's difficult to debug due to the Ludown issue mentioned above.</del>
- [x] Test the Phone skill in the emulator end-to-end with authentication using AAD.
- [x] Write documentation for the Phone skill similar to the documentation for the Email skill (https://github.com/microsoft/botframework-solutions/blob/master/docs/reference/skills/productivity-email.md).  The required permissions are `User.Read`, `Contacts.Read`, and `People.Read`.
- [x] Replace any remaining placeholder responses, such as "Enter your intro message here"
* Nice to have (will be added in later PR's):
  * Add support for free-form phone number types in LUIS
  * Share `NumberNormalizer` with Automotive skill to normalize spelled-out phone numbers
    * Move `NumberNormalizer` to https://github.com/microsoft/Recognizers-Text/tree/master/.NET
  * Add more variety to the responses, so it sounds less dull
  * Support more standardized phone number types (union of the types supported by Graph and Google API)
  * Send content data in events for selection prompts?
  * German language support (if Marion has time)
- [x] Clean up before the PR is ready to merge:
  * Check warnings
  * Remove unused `using`'s
  * Check style
  * Add copyright & license to all files


ryanisgrig:
When telling a client to dial a contact, it should send an Event Activity back with:
Name: OpenDefaultApp
Value: `tel:phonenumbervalue`

This will enable clients (like the Android app we are developing) to use the URI schema and open with their default dialer.

darrenj:
@lauren-mills has made a set of changes to deployment scripts so it would be good to take a copy of the deployment folder from say the Calendar skill and update the deployment resources in your codebase to ensure you have the latest.

skills/src/csharp/phoneskill/phoneskill/Adapters/DefaultAdapter.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.Bot.Builder.Solutions.Middleware;
+using Microsoft.Bot.Builder.Solutions.Responses;
+using Microsoft.Bot.Connector.Authentication;
+using Microsoft.Bot.Schema;
+using PhoneSkill.Responses.Shared;
+using PhoneSkill.Services;
+
+namespace PhoneSkill.Adapters
+{
+    public class DefaultAdapter : BotFrameworkHttpAdapter
```
darrenj, line 17: @lzc850612 made a set of changes to our Skill Template in the last week to remove the HttpAdapter, if you review the Skill template or one of the other skills you should be able to see the (small) changes required

skills/src/csharp/phoneskill/phoneskill/PhoneSkill.csproj
```diff
@@ -0,0 +1,91 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Google.Apis" Version="1.36.1" />
+    <PackageReference Include="Google.Apis.People.v1" Version="1.25.0.830" />
+    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.5.1" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.2.2" />
+    <PackageReference Include="Microsoft.AspNetCore.All" />
+    <PackageReference Include="Microsoft.Azure.CognitiveServices.Language" Version="1.0.1-preview" />
+    <PackageReference Include="Microsoft.Bot.Builder" Version="4.4.3" />
```
darrenj: Update to the latest 4.5 release once your closer to merging this (recently released). Same for the .skills and .solutions versions

darrenj (changes requested):

outofthecave:
@ryanlengel Sending back a `tel:` URI would indeed be convenient for the client, but unfortunately, it's not possible to reliably construct one in the bot.

Converting a given phone number to RFC 3966 format (the format required for `tel:` URI's) is a non-trivial task and requires knowing either the country code of the phone number to be dialled or the default country code that the user's phone currently defaults to.  Given that a lot of users enter phone numbers without country codes into their address book and given that we don't have access to the device that will end up making the call, we can't rely on having enough information to construct an RFC 3966 compliant `tel:` URI.  We could construct the `tel:` URI only in those cases where we have enough information to construct it, but I doubt that is very useful for the client if they can't rely on the URI being present.

I would recommend to leave this task up to the client because the client will most likely be (or at least have access to) the device that will make the actual call.  If that device indeed requires a `tel:` URI, the client can obtain one using the unformatted phone number that we return as part of the final event and a phone number formatting library, for example [libphonenumber](https://github.com/google/libphonenumber).

darrenj:
Updated to target the new `next` branch

darrenj:
@outofthecave Do you have some time to close out the comments and we can get this merged?

outofthecave:
@darrenj I think I already addressed all the comments.  But the PR had gotten unmergeable, so I just rebased it to make it mergeable again.  Can you please have another look and update your review?  Thanks!

darrenj (approved):

Issue #1598, opened 2019-06-19
Remove "do you want to hear more" type prompts from Skills


bobokids:
## Project
Skills
C#

## Description
Skill flow update: Remove the “do you want to hear more” type prompt after every productivity interaction

## To Reproduce

Using calendar skill to get summary of todays' meeting, by end of the conversation, a prompt ask user "Do you want to hear the content of first event" and user has to answer yes or no.

## Expected behavior

Change the prompt to informational msg and allow user to switch to other dialogs.

ryanisgrig:
This also includes the intent switching capability where a Skill will fall back to the VA for processing if it can't handle a user's utterance

PR #1697, opened 2019-07-02
[Part 1, lib change] Add fallback event support

DingmaomaoBJTU:
[Part 1] #1598 Remove "do you want to hear more" type prompts from Skills

## Purpose
Add lib to support *Enable intent switch between skills*,  like the A part in yellow:
image

## Changes
image

*Include illustrative screenshots for context if applicable.*
image

## Tests

## Feature Plan


lib/csharp/microsoft.bot.builder.solutions/microsoft.bot.builder.solutions/FallbackEvents.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Bot.Builder.Solutions
+{
+    public class FallbackEvents
+    {
+        public const string FallbackEventName = "fallback/request";
```
darrenj: @lzc850612  This could be an opportunity to introduce the HandOff ActivityType for this scenario?

lzc850612: Hmmm not sure. This is more like the skill saying 'i don't understand this and maybe you want to take a shot at what the user said' to the calling bot. Handoff doesn't feel right for this case.

lzc850612:
Synced up with Qiong offline on this. There's a few improvements coming to this PR.

darrenj:
@lauren-mills I’ve been thinking about this change which requires template-changes and means manual changes to people who have already created Virtual Assistants. Could you have a think about whether this could be implemented as part of RouterDialog in Bot.Solutions thus meaning we can remove/limit changes to the template MainDialog. I’ve not looked into it in more detail but would be nice if possible.

lzc850612:
> @lauren-mills I’ve been thinking about this change which requires template-changes and means manual changes to people who have already created Virtual Assistants. Could you have a think about whether this could be implemented as part of RouterDialog in Bot.Solutions thus meaning we can remove/limit changes to the template MainDialog. I’ve not looked into it in more detail but would be nice if possible.

i think we can go for that direction as much as possible. if minimum code changes is happening in the VA template it's great. but it's something that has to be done in the VA instead of in a lower level library, then that's that. as for the RouterDialog change, the way i think of it now is that because the SkillDialog has to play a role in this to ask outer dialog (MainDialog from the VA) for another dispatch run (and perhaps in the future when we support full Context Switching we need to add another layer to be able to switch dialogs between skills), it can't be fully done in RouterDialog, not with the Solutions package expanding its role as the lowest level lib that should still work without the skills lib. 

right now after discussion with Qiong, we decided to pass in an additional parameter IIntentRecognizer for the constructor of the skill dialog so that skill dialog can use that to run dispatch again. this way we enable skill dialog to run dispatch again if receiving Fallback (or whatever event we think is right) from the skill. this seems minimum to me as there's no change on MainDialog.cs, and it's probably the right thing to do in terms of what should be the skilldialog's responsibility and the maindialog of the calling VA.

darrenj:
Agreed - we just need to balance the impact on the 5K developers and 3+ deployed VAs with any decision we make moving forwards. It's OK to introduce a "code change" to a template but need to ensure we provide clear documentation on how to enable a new feature as we go.

DingmaomaoBJTU:
Sure~ As we had published our product, the downward compatibility is very important and the impact for old developers should be considered. Currently. fallback event support is optional for old developers and even if they don't change anything in the template it will keep previous behavior. I will add docs to describe this new feature.

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
```diff
@@ -215,25 +291,57 @@ private async Task<DialogTurnResult> ForwardToSkillAsync(DialogContext innerDc,
...
-						dialogResult = await ForwardToSkillAsync(innerDc, _queuedResponses.Dequeue());
-					}
+                    // if there's any response we need to send to the skill queued
+                    // forward to skill and start a new turn
+                    while (_queuedResponses.Count > 0 && dialogResult.Status != DialogTurnStatus.Complete && dialogResult.Status != DialogTurnStatus.Cancelled)
+                    {
+                        if (_queuedResponses.Last().Name == FallbackEvents.FallbackHandleEventName)
+                        {
+                            var lastEvent = _queuedResponses.Dequeue();
+                            if (_skillIntentRecognizer.RecognizeSkillIntentAsync != null)
```
lzc850612: check if _skillIntentRecognizer is null before using it because it could be empty (if the developer didn't specify this argument when instantiating SkillDialog instance). and the behavior should be:
if skillIntentRecognizer specified, run the recognizer
    if the result is an actual intent other than the current skill, launch the confirm dialog to eventually switch to a different skill
    if the result is the same as the current intent, re-send it to the current skill (but this indicates there's discrepency between the VA's dispatch luis model and the skill's luis model, we should log this)
    if the result is empty which means no intent, re-send it to the current skill

@Darrenj does this sound right to you?
     #Resolved

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/ISkillIntentRecognizer.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.Bot.Builder.Dialogs;
+using Microsoft.Bot.Builder.Skills.Models.Manifest;
+
+namespace Microsoft.Bot.Builder.Skills
+{
+    public interface ISkillIntentRecognizer
+    {
+        Func<DialogContext, Task<SkillManifest>> RecognizeSkillIntentAsync { get; }
```
lzc850612: it seems like only the id property of the return value of this Func is being used so why not just make the return type string? #Resolved

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
```diff
@@ -215,25 +291,57 @@ private async Task<DialogTurnResult> ForwardToSkillAsync(DialogContext innerDc,
...
-					// forward to skill and start a new turn
-					while (_queuedResponses.Count > 0 && dialogResult.Status != DialogTurnStatus.Complete && dialogResult.Status != DialogTurnStatus.Cancelled)
-					{
-						dialogResult = await ForwardToSkillAsync(innerDc, _queuedResponses.Dequeue());
-					}
+                    // if there's any response we need to send to the skill queued
+                    // forward to skill and start a new turn
+                    while (_queuedResponses.Count > 0 && dialogResult.Status != DialogTurnStatus.Complete && dialogResult.Status != DialogTurnStatus.Cancelled)
+                    {
+                        if (_queuedResponses.Last().Name == FallbackEvents.FallbackHandleEventName)
```
lzc850612: define a variable here to take _queuedResponses.last() and use that for checking and calling other functions going forward. #Resolved

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
```diff
@@ -215,25 +291,57 @@ private async Task<DialogTurnResult> ForwardToSkillAsync(DialogContext innerDc,
...
+
+                                        return await innerDc.BeginDialogAsync(DialogIds.ConfirmFlow, options);
+                                    }
+
+                                    await _skillTransport.CancelRemoteDialogsAsync(innerDc.Context);
+                                    return await innerDc.EndDialogAsync(recognizedSkillManifestRecognized.Id);
+                                }
+                            }
+
+                            dialogResult = await ForwardToSkillAsync(innerDc, lastEvent);
```
lzc850612: what's the point of sending the fallback event back to the current skill? i'm assuming this is so that the skill processes user's utterance anyway? how are you going to be able to find that previous activity now that this is essentially another turn?

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
```diff
@@ -29,36 +32,88 @@ public class SkillDialog : ComponentDialog
...
+            }
+
+            return await sc.NextAsync();
+        }
+
+        public async Task<DialogTurnResult> FinishIntentSwitch(WaterfallStepContext sc, CancellationToken cancellationToken)
+        {
+            if (sc.Options != null && sc.Options is SkillSwitchConfirmOption skillSwitchConfirmOption)
+            {
+                if (sc.Result is bool result && result == true)
```
lzc850612: "== true"  is not needed. #Resolved

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
```diff
@@ -215,25 +291,57 @@ private async Task<DialogTurnResult> ForwardToSkillAsync(DialogContext innerDc,
...
+                                    if (_skillIntentRecognizer.ConfirmIntentSwitch)
+                                    {
+                                        var options = new SkillSwitchConfirmOption()
+                                        {
+                                            LastActivity = lastEvent,
+                                            TargetIntent = recognizedSkillManifestRecognized.Id,
+                                            UserInputActivity = innerDc.Context.Activity,
+                                        };
+
+                                        return await innerDc.BeginDialogAsync(DialogIds.ConfirmFlow, options);
```
lzc850612, line 349: so after the users says 'no, don't switch to a different skill', shouldn't we send the original activity back to the original skill? seems like you're just simply doing a return here for all cases. should get the result of the ConfirmFlow and use that to drive what happens next.

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialogOptions.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Microsoft.Bot.Builder.Skills
+{
+    using System;
+    using System.Threading.Tasks;
+    using Microsoft.Bot.Builder.Dialogs;
+
+    public class SkillDialogOptions
```
lzc850612: where is this being used? #Resolved

lib/csharp/microsoft.bot.builder.solutions/microsoft.bot.builder.solutions/Dialogs/RouterDialog.cs
```diff
@@ -66,10 +66,19 @@ protected override async Task<DialogTurnResult> OnContinueDialogAsync(DialogCont
 
                                     case DialogTurnStatus.Complete:
                                         {
-                                            await CompleteAsync(innerDc).ConfigureAwait(false);
+                                            if (result.Result is string)
+                                            {
+                                                // Redispatch
+                                                await RedispatchAsync(innerDc, result).ConfigureAwait(false);
```
lzc850612: why do we still need the redispatch here?

lib/csharp/microsoft.bot.builder.solutions/microsoft.bot.builder.solutions/FallbackEvents.cs
lzc850612: we should add the events here in the file models/SkillEvents.cs because that's where all the events needed for skill are defined. they should also start with 'skill/' so it's clear that they're for skills protocol #Resolved

lib/csharp/microsoft.bot.builder.solutions/microsoft.bot.builder.solutions/Resources/CommonStrings.Designer.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.Bot.Builder.Solutions.Resources {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "16.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "15.0.0.0")]
     [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
```
lzc850612: are you using VS 2017 instead of 2019 which is where this change is coming from? #Resolved

lib/csharp/microsoft.bot.builder.solutions/microsoft.bot.builder.solutions/BotSettingsBase.cs
```diff
@@ -99,6 +99,14 @@ public class BotSettingsBase
         /// </value>
         public Dictionary<string, string> Properties { get; set; } = new Dictionary<string, string>();
 
+        /// <summary>
+        /// Gets or sets whether to prompt a confirm dialog when do intent switch.
+        /// </summary>
+        /// <value>
+        /// The confirm intent switch setting.
+        /// </value>
+        public bool ConfirmIntentSwitch { get; set; }
```
lzc850612: not sure if we want to add this here at the Microsoft.Bot.Builder.Solutions level. this lib should be not dependent on skills meaning that people who use this lib doesn't have to be using skills and in that sense we don't have to have this in here which only work when the VA has skills. at this point let's remove it and keep the flag on the SkillDialog level meaning when instantiating a skilldialog the developers can specify this flag so different skill dialogs can have different flags. we can consider adding a global setting later. #Resolved

lzc850612 (changes requested):
This PR shouldn't be merged into master as we don't want this to be in our very next release. Can you create the PR against 4.future branch later on after people are done with the review since that's already started? thanks.

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
DingmaomaoBJTU: Yes, I missed that~ Add null check and related comments.

---
In reply to: [301378037](https://github.com/microsoft/botframework-solutions/pull/1697#discussion_r301378037) [](ancestors = 301378037)

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialogOptions.cs
DingmaomaoBJTU: This should be deleted~

---
In reply to: [301401438](https://github.com/microsoft/botframework-solutions/pull/1697#discussion_r301401438) [](ancestors = 301401438)

lib/csharp/microsoft.bot.builder.solutions/microsoft.bot.builder.solutions/BotSettingsBase.cs
DingmaomaoBJTU: So developers can only set this flag in code, right?

---
In reply to: [301402835](https://github.com/microsoft/botframework-solutions/pull/1697#discussion_r301402835) [](ancestors = 301402835)

lzc850612: yes at least that's something we should start with #Resolved

lib/csharp/microsoft.bot.builder.solutions/microsoft.bot.builder.solutions/Resources/CommonStrings.Designer.cs
DingmaomaoBJTU: Yes.... Let me update my VS later.

---
In reply to: [301402216](https://github.com/microsoft/botframework-solutions/pull/1697#discussion_r301402216) [](ancestors = 301402216)

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
DingmaomaoBJTU: Yes, currently this is done in confirm flow. Related information is saved SkillSwitchConfirmOption and it will be passed to confirm flow. This is a multi-turn flow so I'm not sure whether we can get flow result of confirmflow here~

---
In reply to: [301401287](https://github.com/microsoft/botframework-solutions/pull/1697#discussion_r301401287) [](ancestors = 301401287)

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
DingmaomaoBJTU: The fallbackhandle event is used to make skills pause to wait for VA bot's confirmation of fallback handling. As Skill already knew it cannot handle previous activity, so it doesn't need previous activity to continue its next steps. Does this make sense to you? @tedlee@microsoft.com 

---
In reply to: [301380656](https://github.com/microsoft/botframework-solutions/pull/1697#discussion_r301380656) [](ancestors = 301380656)

lib/csharp/microsoft.bot.builder.solutions/microsoft.bot.builder.solutions/Dialogs/RouterDialog.cs
DingmaomaoBJTU: Because the RecognizeSkillIntentAsync() (controled by skilldialog) only can return new intent recognized by VA, but cannot begin a new skill dialog (it is a VA job). So here I call a RedispatchAsync in VA to begin a new skill dialog.

---
In reply to: [301401852](https://github.com/microsoft/botframework-solutions/pull/1697#discussion_r301401852) [](ancestors = 301401852)

DingmaomaoBJTU: protected override async Task RedispatchAsync(DialogContext innerDc, DialogTurnResult result = null, CancellationToken cancellationToken = default(CancellationToken))
        {
            var intent = result.Result as string;

            // Identify if the dispatch intent matches any Action within a Skill if so, we pass to the appropriate SkillDialog to hand-off
            var identifiedSkill = SkillRouter.IsSkill(_settings.Skills, intent.ToString());

            if (identifiedSkill != null)
            {
                // We have identiifed a skill so initialize the skill connection with the target skill
                var dialogResult = await innerDc.BeginDialogAsync(identifiedSkill.Id);

                if (dialogResult.Status == DialogTurnStatus.Complete)
                {
                    await CompleteAsync(innerDc);
                }
            }

            await base.RedispatchAsync(innerDc, result, cancellationToken);
        }

---
In reply to: [301872078](https://github.com/microsoft/botframework-solutions/pull/1697#discussion_r301872078) [](ancestors = 301872078,301401852)

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/ISkillIntentRecognizer.cs
DingmaomaoBJTU: Yes, that looks better~

---
In reply to: [301378859](https://github.com/microsoft/botframework-solutions/pull/1697#discussion_r301378859) [](ancestors = 301378859)

DingmaomaoBJTU:
> This PR shouldn't be merged into master as we don't want this to be in our very next release. Can you create the PR against 4.future branch later on after people are done with the review since that's already started? thanks.

- Sure, will do after sign off.

lib/csharp/microsoft.bot.builder.solutions/microsoft.bot.builder.solutions/Dialogs/RouterDialog.cs
lzc850612: in RouteAsync method of the MainDialog.cs, we can do this:

            // Identify if the dispatch intent matches any Action within a Skill if so, we pass to the appropriate SkillDialog to hand-off
            var identifiedSkill = SkillRouter.IsSkill(_settings.Skills, intent.ToString());

            if (identifiedSkill != null)
            {
                // We have identiifed a skill so initialize the skill connection with the target skill
                var result = await dc.BeginDialogAsync(identifiedSkill.Id);

                // when the skill returns with an intent we need to invoke the skill with the intent
                // until there's no intent from the skill dialog
                while (result.Result is string intentResult && !string.IsNullOrWhiteSpace(intentResult))
                {
                    // Identify if the dispatch intent matches any Action within a Skill if so, we pass to the appropriate SkillDialog to hand-off
                    identifiedSkill = SkillRouter.IsSkill(_settings.Skills, intent.ToString());

                    if (identifiedSkill != null)
                    {
                        // We have identiifed a skill so initialize the skill connection with the target skill
                        result = await dc.BeginDialogAsync(identifiedSkill.Id);
                    }
                    else
                    {
                        break;
                    }
                }

                if (result.Status == DialogTurnStatus.Complete)
                {
                    await CompleteAsync(dc);
                }
            }

this will use the result of the skill dialog and immediately decide to launch a different skill dialog if needed. This way we don't have to do redispatch in the lower level lib.

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
lzc850612: well you return different stuff in the FinishIntentSwitch method for yes/no. you should just get that here, and then use that to drive what happens next, if yes then EndDialog with the targetIntent, if not then resend the activity back to the same skill. right now you're always cancelling the remote skill dialog and ending the current skill dialog

lzc850612 (changes requested):

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
DingmaomaoBJTU: Sure, I move resend logic to FinishIntentSwitch() and add related comments. And one additional logic I add here to make it work is I need to manually change DialogTurnResult from DialogTurnStatus.Complete to DialogTurnStatus.Waiting when user cancel skill switching. otherwise, the SkillDialog will end~

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/DispatchDialog.cs
```diff
@@ -0,0 +1,120 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Bot.Builder.Dialogs;
+using Microsoft.Bot.Builder.Solutions.Dialogs;
+using Microsoft.Bot.Builder.Solutions.Extensions;
+using Microsoft.Bot.Schema;
+
+namespace Microsoft.Bot.Builder.Skills
+{
+    public class DispatchDialog : RouterDialog
```
lzc850612: let's create a 'Dialogs' folder and put this class under that folder.

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/DispatchDialog.cs
```diff
@@ -0,0 +1,120 @@
...
+                                {
+                                    case DialogTurnStatus.Empty:
+                                        {
+                                            await RouteAsync(innerDc).ConfigureAwait(false);
+                                            break;
+                                        }
+
+                                    case DialogTurnStatus.Complete:
+                                        {
+                                            if (result.Result is string)
```
lzc850612: let's define a class that's called Intent (or something better) and check the Result property against that type and use it. 'String' as a type is too general and it'll collide with something else in the future easily.

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
```diff
@@ -268,5 +388,29 @@ private Action<Activity> GetTokenRequestCallback(DialogContext dialogContext)
...
         }
+
+        private Action<Activity> GetFallbackCallback(DialogContext dialogContext)
+        {
+            return (activity) =>
+            {
+                // Send trace to emulator
+                dialogContext.Context.SendActivityAsync(new Activity(type: ActivityTypes.Trace, text: $"<--Received a fallback request from a skill")).GetAwaiter().GetResult();
+
+                var fallbackHandleEvent = activity.CreateReply();
```
lzc850612: the variable name should be fallbackEvent, instead of fallbackHandleEvent since we're not declaring that here.

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillSwitchConfirmOption.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.Bot.Schema;
+
+namespace Microsoft.Bot.Builder.Skills
+{
+    public class SkillSwitchConfirmOption
+    {
+        public Activity LastActivity { get; set; }
```
lzc850612: let's just call LastActivity 'FallbackHandledEvent'

lib/csharp/microsoft.bot.builder.solutions/microsoft.bot.builder.solutions/Dialogs/RouterDialog.cs
```diff
@@ -138,6 +139,18 @@ protected virtual Task CompleteAsync(DialogContext innerDc, DialogTurnResult res
...
         }
 
+        /// <summary>
+        /// Called when fallbackhandler event is recieved.
+        /// </summary>
+        /// <param name="innerDc">The dialog context for the component.</param>
+        /// <param name="result">The dialog result when inner dialog completed.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        protected virtual Task RedispatchAsync(DialogContext innerDc, DialogTurnResult result = null, CancellationToken cancellationToken = default(CancellationToken))
```
lzc850612: changes in this file are not needed any more.

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillWebSocketBotAdapter.cs
```diff
@@ -232,6 +233,17 @@ public async Task SendRemoteTokenRequestEventAsync(ITurnContext turnContext, Can
             await SendActivitiesAsync(turnContext, new Activity[] { response }, cancellationToken).ConfigureAwait(false);
         }
 
+        public async Task SendRemoteFallbackEventAsync(ITurnContext turnContext, CancellationToken cancellationToken)
+        {
+            // We trigger a Fallback Request from the Parent Bot by sending a "FallbackRequest" event
```
lzc850612: This function needs to be defined in an interface so the skill dialogs can get the adapter and cast it into the interface and use it. it can't just be a public method

the comment in line 238 should be 'by sending a 'skill/fallbackRequest' event'

lzc850612 (changes requested):

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/DispatchDialog.cs
DingmaomaoBJTU: Sure, good suggestion!

lzc850612:
Let's use the branch 'future' for merging of this PR, once we're done reviewing

Issue #1663, opened 2019-06-25
Point of Interest Skill - Forward Fill to get directions


darrenj:
# Simplify conversation flow to get directions after finding a PoI

Today in the Point of Interest skill you identify a PoI and then go through a series of prompts to get driving directions to a PoI, for example - see the multiple responses below:

![image](https://user-images.githubusercontent.com/33519520/60140884-0652e680-9768-11e9-9cfc-de0594f0da7a.png)

**Describe the solution you'd like**

Enabling the user to shortcut subsequent prompts by saying "drive me to the first one" would enable us to shortcut the following prompts. The user can still say "first one" and get the prompt but could shortcut if they ask?

![image](https://user-images.githubusercontent.com/33519520/60140858-f0452600-9767-11e9-986e-0f32956b0acb.png)

ryanisgrig:
The Azure Maps API has the potential to return an array of potential routes to a destination, but in all of my testing I have only experienced a single route returned. Currently this results in a ConfirmPrompt to ensure a user wants that route but I think if there is a single route returned that should be considered the "shortcut" and instead POI just sends the event to trigger a client to open a map (see #1443 for how the event should be configured)

PR #1713, opened 2019-07-03
use geographyV2 and submit action in poi

### LINKED ISSUE #1587: Adaptive Cards in Skills need to be clickable when relevant
**Is your feature request related to a problem? Please describe.**

In #1439 we are enabling the Adaptive Cards from a VA to be clickable, on a tablet this gives the user to not only need to use voice input to navigate through their conversation. 

1. Investigate each Skill where Adaptive Cards are displayed and determine if it makes sense to have a card be clickable. _For example, in the Calendar Skill it may not make sense to have the overview card be clickable as it is unclear what the next step of a conversation would be there._
1. If it is determined that the card should be clickable, implement the correct Action property to the card.
    * In the Point of Interest Skill, selecting a card should send an [Action.Submit](https://adaptivecards.io/explorer/Action.Submit.html) with the appropriate data to trigger the next step of the flow. In this case, Action.Submit may have the POI's name in the data property. This data is consumed by the VA as a selection and it navigates to the next step in the dialog.
image

**Additional context**

@Dewain27 to add more context

### LINKED ISSUE #1662: Update PoI skill to leverage Address and geography entities
The Point of Interest skill makes use of keyword entities to find types of point of interest but doesn't make use of the `Address` and `geographyv2` entities which the LUIS model provides.

For example each of the following questions should not prompt for location (which they currently do). When calculating routing directions we do need to know the users current location which in most cases will have been provided through a location event prior to this. I'm OK at this stage for us to not offer this capability if location hasn't been provided to simplify the interaction.

- `Find a coffee shop in Seattle`

In this scenario the Skill would go straight to providing coffee shops in Seattle. If a "location" has been provided (through the event) this is **not** used to find a PoI (which it used for today) but **will** be used to calculate a route to the identified coffee shop from the provided location. If a location has not been provided we should not offer to provide routing directions (we could explore prompting for current location?). Therefore we do a straight forward poi search in seattle and provide routing from the provided location if available.

image

- `Get Directions to Mt Rainier`
In this scenario the Skill would find Mt Rainier PoI (ideally not prompt for multiple options if that works) and then as above provide routing options if the location event has sent location information.
image

- `I need to go to Seatac`
Same as above
image

- `I want to go to 900 bellevue way ne, Bellevue`
In this case we need to build up the address from the address entities and then as above providing routing if we have a location.
image

**Describe alternatives you've considered**

**Additional context**

xieofxie:
## Purpose

This closes #1662, closes #1587.

## Changes

* To avoid overwrite address and keyword, PoI LUIS is used only once at the beginning(RouteAsync) since it is not meaningful to parse responses afterwards
* Use a combination of geographyV2_poi and geographyV2_city when address is empty
* When click on card, submit confirm as ConfirmPrompt for single choice, value of suggested action for multiple choices. However, it will always submit the content when clicked
* use SelectPointOfInterestPrompt for multiple routes
* Skip event ActiveRoute.Directions in Teams

## Tests

selectAction is not working in Teams.

## Feature Plan

N/A.


darrenj (approved):

lzc850612:
@ryanlengel can you take a look at this PR?

ryanisgrig:
Why do we want geographyv2 and ADDRESS? Shouldn't geographyv2 functionally overwrite the need for a custom ADDRESS keyword?

Or is it just as a fallback for when ADDRESS isn't picked up?

darrenj:
Address takes priority and fallback is geography, see bugs I raised last week for examples

ryanisgrig:
Great, don't merge yet. I can validate later today :)

xieofxie:
> Why do we want geographyv2 and ADDRESS? Shouldn't geographyv2 functionally overwrite the need for a custom ADDRESS keyword?
> 
> Or is it just as a fallback for when ADDRESS isn't picked up?

I didn't figure out why it uses some very long phraseLists instead of geographyv2 and ADDRESS sometimes does better than geographyv2. So I use geographyv2 as a fallback.

ryanisgrig:
If a single card is displayed and the confirm (yes/no) prompt is shown, I think that a user could still be expected to click on it there (if they are used to clicking on others). Thus they would either click or say yes to have to select it. I'm assuming in this case the select action is broken on this prompt for the "submit empty" issue you linked to?

xieofxie:
> If a single card is displayed and the confirm (yes/no) prompt is shown, I think that a user could still be expected to click on it there (if they are used to clicking on others). Thus they would either click or say yes to have to select it. I'm assuming in this case the select action is broken on this prompt for the "submit empty" issue you linked to?

No, submit yes is fine. I compared two submit behaviors: nothing(only 1) + name(>1) with yes(only 1) + name(>1) and favor the first one. However, submit empty is broken and I didn't figure out a way to remove the Action easily.
Now I am working on [submit in Teams](https://github.com/OfficeDev/BotBuilder-MicrosoftTeams/issues/191) and we could merge this later.

xieofxie:
Now I copied code from ConfirmPrompt to submit Yes for single card. But Ideally it may use it directly (#1751).
Teams only supports actions not selectAction and it will introduce a button in the card, so I don't include it.

xieofxie:
I added a multiple prompt for routes and I would like to know how to handle other languages for responses (use some online translator?).

xieofxie:
* Codes from ConfirmPrompt are copied to submit 'Yes' for single card. But Ideally these should be used directly (#1751).
* Teams only supports actions not selectAction and it will introduce a button in the card, so it is not included.
* Prompt for multiple routes is added with multiple language support by Microsoft translator.
* After this pr, #1663 could be easily solved with some updates of tests.

Apology: I shouldn't edit comments which make it very hard for reviewing..

xieofxie:
Use CultureInfo.CurrentUICulture instead of Activity.Locale.

ryanisgrig (approved):

ryanisgrig:
Can you fix the merge conflict? Then we can close, thanks!

xieofxie:
Merge conflict fixed.

PR #1722, opened 2019-07-03
Updates skill deployment scripts with latest fixes from VA template

lauren-mills:
## Purpose
- Update deployment scripts to be consistent between VA and skill projects
- Fix issues with update_cognitive_models.ps1 in skill project

## Changes
- removed add_remote_skill.ps1 and supporting scripts from VA template

## Tests
N/A

## Feature Plan
N/A


templates/Skill-Template/csharp/Sample/SkillSample/Deployment/Scripts/update_cognitive_models.ps1
```diff
@@ -39,21 +39,42 @@ foreach ($langCode in $languageMap.Keys) {
...
             # If the LUIS application id already exists within the model no action will be taken
-            Write-Host "> Adding $($luisApp.id) app to dispatch model ... "
-			(dispatch add `
-				--type "luis" `
-				--name $luisApp.name `
-				--id $luisApp.appid  `
-				--intentName "l_$($luisApp.id)" `
-                --dispatch $(Join-Path $dispatchFolder $langCode "$($dispatch.name).dispatch") `
-                --dataFolder $(Join-Path $dispatchFolder $langCode))  2>> $logFile | Out-Null
+			if ($dispatch) {
```
lzc850612, line 67: guess we're just using $dispatch for a placeholder for now so we can add it in the future?

lzc850612 (approved):

templates/Skill-Template/csharp/Sample/SkillSample/Deployment/Scripts/update_cognitive_models.ps1
lauren-mills: In this script its looking at what's in the cognitivemodels.json file, so if they have added a dispatch model it would update it

lauren-mills: In the deploy_cognitive_models script is where I have the $useDispatch flag that can be set to create a dispatch model or not

ryanisgrig:
Can you enumerate what issues were in `update_cognitive_models.ps1`?

lauren-mills:
The issues were that it was expecting there to be a dispatch model and knowledgebases so I made that optional, and there was an issue with it not parsing the exported .luis files into .lu files when the -RemoteToLocal flag is provided. So in that case, it wasn't updating the local files correctly

xieofxie:
It seems that template.json in Template is not updated as the one in SkillSample.
Do we need to copy these changes to skills? Or should we provide some tools to automatically sync changes from templates to skills?

PR #1726, opened 2019-07-03
Update POI deployment scripts

lauren-mills:
## Purpose
Update POI scripts to be consistent with skill template and va

## Changes
- updated deployment scripts with changes from #1722 and #1694
- Fixed camel casing of lu file name, luisgen file, and ids in cognitivemodels.json
- moved azure maps and other additional config out of "properties" in appsettings (@ryanlengel - take a look at this and let me know what you think)

## Tests
N/A

## Feature Plan
N/A


skills/src/csharp/pointofinterestskill/pointofinterestskill/appsettings.json
```diff
@@ -1,27 +1,29 @@
...
     "cosmosDBEndpoint": "",
-    "authKey": ""
+    "collectionId": "botstate-collection",
+    "databaseId": "botstate-db"
+  },
+  "luis": {
+    "accountName": "",
+    "key": ""
   },
-  "properties": {
```
ryanisgrig, line 19: The other Skills use a properties object for their additional dependencies and settings, I think this should be reverted

skills/src/csharp/pointofinterestskill/pointofinterestskill/appsettings.json
```diff
@@ -1,27 +1,29 @@
 {
   "microsoftAppId": "",
   "microsoftAppPassword": "",
-  "oauthConnections": [
-  ],
-  "ApplicationInsights": {
+  "botWebAppName": "",
```
ryanisgrig, line 4: Will all these new settings be required for a skill to run correctly? i.e. will it break the daily build if these are left empty?

ryanisgrig (changes requested):

xieofxie:
update_cognitive_models.ps1 with RemoteToLocal will complain 'Unrecognized LUIS locale'. Should use `luis get application` to get full length culture.

skills/src/csharp/pointofinterestskill/pointofinterestskill/appsettings.json
lauren-mills: No, I just changed the script to write all arm template outputs to appsettings so that all the customizations (like the azure maps key) don’t need any custom stuff in the script. The code doesn’t use them though. Only the stuff that’s included in the botsettings.cs file will attempt to be deserialized, and even then if it’s not there it will just be null

skills/src/csharp/pointofinterestskill/pointofinterestskill/appsettings.json
lauren-mills: I’m going to change them all. The only reason we needed a propeties object was because of in proc. Since we aren’t trying to make the settings exactly conform between the skills, this is more conventional

lauren-mills:
@xieofxie I changed the deploy_cognitive_models script to write the culture to the cognitivemodels.json file when they are deployed to make sure its present for the update, but I’ll look into Luis get application as well

skills/src/csharp/pointofinterestskill/pointofinterestskill/appsettings.json
lauren-mills: Plus, I can’t easily write out the azureMaps key into the properties property without making the script custom for POI

ryanisgrig (approved):

ryanisgrig:
I've added an internal note for me to update all the skill builds after this & similar PRs. The solutions aren't hooked up to their rings yet so it won't affect anything published.

PR #1794, opened 2019-07-10
poi shows single route directly (#1663)

### LINKED ISSUE #1663: Point of Interest Skill - Forward Fill to get directions

xieofxie:
## Purpose

Close #1663.

## Changes

*Include illustrative screenshots for context if applicable.*

* No prompt for single route
* CancelRouteFailureTest is not valid now and is removed
* SingleRouteFound is modified and translated by Bing..

## Tests

Yes.

## Feature Plan


ryanisgrig:
This looks great and is a much better experience!

image

skills/src/csharp/pointofinterestskill/pointofinterestskill/Dialogs/MainDialog.cs
```diff
@@ -242,7 +243,7 @@ protected override async Task<InterruptionAction> OnInterruptDialogAsync(DialogC
         {
             var result = InterruptionAction.NoAction;
 
-            if (dc.Context.Activity.Type == ActivityTypes.Message)
+            if (dc.Context.Activity.Type == ActivityTypes.Message && !string.IsNullOrEmpty(dc.Context.Activity.Text))
```
ryanisgrig, line 246: What message activities are missing text? Just wondering.

ryanisgrig (approved):

skills/src/csharp/pointofinterestskill/pointofinterestskill/Dialogs/MainDialog.cs
xieofxie: The SubmitText is empty for single route and if one clicks on it, it will send an empty text which will make LUIS return an empty Intent. Related [issue](https://github.com/microsoft/botbuilder-dotnet/issues/2174).

xieofxie:
Update for conflicts.

lauren-mills (approved):

PR #1870, opened 2019-07-16
Fix Skill app settings from refactor

ryanisgrig:
### Purpose
Other skills have removed the `properties` object and moved those keys to the root-level, this does the same for the Search and Auto Skills.

### Changes
N/a

### Tests
N/a

### Feature Plan
N/a


lauren-mills (approved):

lzc850612 (approved):

ryanisgrig:
Recycled exposed keys

PR #1927, opened 2019-07-23
Updated to IBot Implementation across projects

### LINKED ISSUE #1599: Implement ActivityHandler correctly in DialogBot
**Is your feature request related to a problem? Please describe.**
DialogBot inherits from ActivityHandler but the OnTurnAsync method never calls it's base member. There is also the RouterDialog inherited by Main which appears to be doing much the same job as the AcitivityHandler.

Either inherit DialogBot from IBot and cut out the unused inheritence or remove the RouterDialog and shift the bulk of the logic from Main into a _<VirtualAssistant>_Bot class.

**Describe alternatives you've considered**
Or even remove the router/handler element entirely and allow the active dialog to handle events in context such as responses to adaptive cards.

**Additional context**
Currently both the ActivityHandler and RouterDialog fail silentliy for unsupported Activities. This is particularly an issue around Teams Authentication where the events are never reaching the appropriate OAuthPrompt.

lauren-mills:
close #1599 

### Purpose
As mentioned in the issue, the previous implementation did not use the full features of activityHandler since we handle this logic in RouterDialog. We should readdress implementing ActivityHandler and deprecating routerDialog in future releases.

### Changes
- Changed from ActivityHandler to IBot in DialogBot.cs
- Updated dialog stack logic to latest SDK implementation using Dialog.RunAsync()

### Tests
-N/A

### Feature Plan
- We should readdress implementing ActivityHandler and deprecating routerDialog in future releases.

### DIFF
--- a/skills/src/csharp/automotiveskill/automotiveskill/Bots/DialogBot.cs
+++ b/skills/src/csharp/automotiveskill/automotiveskill/Bots/DialogBot.cs
@@ -11,23 +11,23 @@ using Microsoft.Extensions.DependencyInjection;
 
 namespace AutomotiveSkill.Bots
 {
-    public class DialogBot<T> : ActivityHandler
+    public class DialogBot<T> : IBot
         where T : Dialog
     {
+        private readonly Dialog _dialog;
+        private readonly BotState _conversationState;
+        private readonly BotState _userState;
         private readonly IBotTelemetryClient _telemetryClient;
-        private DialogSet _dialogs;
 
         public DialogBot(IServiceProvider serviceProvider, T dialog)
         {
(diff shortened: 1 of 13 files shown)


lzc850612:
Do we want this to go to this week's release?

lauren-mills:
@lzc850612 I think it is minor enough that we should. Also, since we're on 4.5.1 sdk now, we should implement their approach to prevent confusion.

lzc850612 (changes requested):
should we remove the autosavestatemiddleware now that we're doing it explicitly?

darrenj (approved):

lzc850612 (approved):

PR #1948, opened 2019-07-25
Use consistent header name for authorization header

### LINKED ISSUE #1947: TypeScript - Skill authentication broken
botframework-solutions

TypeScript

#### What happens?
When trying to enable authentication on a TypeScript skill, the authentication always fails.
This seems to be because the SkillHttpAdapter in the skill is looking for the JWT token in a header called "Authorization" but the VA is sending the JWT in a header called "authorization" (note the capitalisation)

Implement a skill using TypeScript. Enable authentication.

henryjenkins:
### Purpose
Close #1947: Use the same capitalisation for authorization header to ensure that the VA can talk to an authenticated TypeScript skill

### DIFF
--- a/lib/typescript/botbuilder-skills/src/http/skillHttpAdapter.ts
+++ b/lib/typescript/botbuilder-skills/src/http/skillHttpAdapter.ts
@@ -1,3 +1,4 @@
+import { Constants } from '@azure/ms-rest-js';
 import { BotFrameworkAdapter, BotFrameworkAdapterSettings, BotTelemetryClient, InvokeResponse,
     Severity, TurnContext, WebRequest, WebResponse } from 'botbuilder';
 import { Activity } from 'botframework-schema';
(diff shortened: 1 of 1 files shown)


dfavretto (changes requested):
Hi @henryjenkins, thanks for contributing!

This change looks good, however, we discussed with the SDK Team and they are removing the `@azure/ms-rest-js` dependency from the SDK, as this library is causing some issues. Here you can see the ongoing work:
- Microsoft/botbuilder-js#1071
- Microsoft/botbuilder-js#1072

Because of this, we are planning to remove this dependency as soon as they finish doing so in the SDK.

If you take a look at the [`botFrameworkAdapter`](https://github.com/microsoft/botbuilder-js/blob/07b6ef9d0508d0f2e36966b8742e1a855ff8d411/libraries/botbuilder/src/botFrameworkAdapter.ts#L582) class, which the `skillHttpAdapter` inherits from, this problem is solved doing this:
```typescript
const authHeader: string = req.headers.authorization || req.headers.Authorization || '';
```
In this way, both lower case and upper case are being considered, so the best approach would be to replicate the `botFrameworkAdapter` implementation.

dfavretto (approved):

darrenj (approved):

Issue #1952, opened 2019-07-25
LUIS and QnA models in VA Template need to reflect new capitalization changes


ryanisgrig:
#### What project is affected?
Virtual Assistant Template
#### What language is this in?
C# 

#### What happens?
The LUIS and QnA properties in DispatchLuis.cs need to reflect new naming scheme

#### What are the steps to reproduce this issue?
Project does not build 

#### What were you expecting to happen?
Project builds

PR #1956, opened 2019-07-25
[VA Template] Fixed additional reference to luis service

ryanisgrig:
### Purpose
Through end to end testing found one more reference to the incorrect General LUIS model (#1952)

### Changes
n/a 

### Tests
n/a 

### Feature Plan

#### Deployment Scripts
- [x] I have replicated my changes in the **Virtual Assistant Template** and **Sample** projects

### DIFF
--- a/templates/Virtual-Assistant-Template/csharp/Sample/VirtualAssistantSample/Dialogs/MainDialog.cs
+++ b/templates/Virtual-Assistant-Template/csharp/Sample/VirtualAssistantSample/Dialogs/MainDialog.cs
@@ -104,7 +104,7 @@ namespace VirtualAssistantSample.Dialogs
 
                 if (luisService == null)
                 {
-                    throw new Exception("The general LUIS Model could not be found in your Bot Services configuration.");
+                    throw new Exception("The General LUIS Model could not be found in your Bot Services configuration.");
                 }
                 else
                 {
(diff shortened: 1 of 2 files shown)


lzc850612:
the lu file is general.lu which is lower case. wouldn't it generate into lowercase model?

lzc850612 (changes requested):
the lu file is general.lu which is lower case. wouldn't it generate into lowercase model?

lzc850612:
> the lu file is general.lu which is lower case. wouldn't it generate into lowercase model?

and if you have to make this change to make the general intents to work, doesn't that mean this never worked before? i don't think that's the case.

ryanisgrig:
I thought the cognitive models I created with the deploy script made them uppercase, or at least that's the json I've been testing with (I don't remember manually changing them). Is that not the case?

lzc850612:
> I thought the cognitive models I created with the deploy script made them uppercase, or at least that's the json I've been testing with (I don't remember manually changing them). Is that not the case?

this is the lu file which is lower case itself. https://github.com/microsoft/botframework-solutions/blob/master/templates/Virtual-Assistant-Template/csharp/Sample/VirtualAssistantSample/Deployment/Resources/LU/en/general.lu

i don't think the deployment script would make it upper case even if the lu file itself is lower case. We did change the lu files for all the skills to be camelcase instead of lower case. @lauren-mills does the deploy script change casing when creating the luis model? should we change the Lu files in the templates so that they're consistent with the skills change?

ryanisgrig:
I found it, my local copy of this repo has these files as uppercase (🤷🏻‍♂️) but I can see here they should be lowercase. Closing this one and reverting my last PR.

PR #2028, opened 2019-08-02
[HospitalitySkill] Adding RequestItems dialog

litofish:
### Purpose
Adding a new feature to the Hospitality Skill that allows a user to request specific items from the hotel.

### Changes
Should handle requests for a certain quantity of an item and any number of items that are requested. 

#### General

#### Bots


lauren-mills (approved):

skills/src/csharp/experimental/hospitalityskill/Data/AvailableItems.json
```diff
@@ -0,0 +1,33 @@
+  [
+    "towel",
+    "towels",
+    "hand towels",
```
darrenj, line 4: You may have already had some discussions about this but you might want to move this to LUIS LIst entities, you can have canonical terms such as (hand towel) and then add synonyms for each item. That way we push the processing into LUIS for that and simplify your code?

litofish: I had used list entities initially, but ran into an issue where LUIS would return multiple entities if the items had similar words in them - saying hand towel would return both hand towel and towel, rather than just hand towel. So I ended up just using a simple entity with a phrase list instead, but not sure this is the best solution either.

darrenj: I'll connect you with Nayer to run the goal past him and see if he can help, otherwise we can go with this (feel free to merge and we can update afterwards if easier)

lzc850612 (approved):

skills/src/csharp/experimental/hospitalityskill/Data/AvailableItems.json
litofish: Ok! I will merge and update after meeting with Nayer.

PR #2138, opened 2019-08-16
Set CallerId in SkillWebSocketTransport

### LINKED ISSUE #2137: SkillWebSocketTransport::CancelRemoteDialogsAsync does not have CallerId

Microsoft.Bot.Builder.Skills

C#

#### What happens?

CallerId is not set when send

xieofxie:
### Purpose

Close #2137 

### Changes

Set CallerId in SkillWebSocketTransport instead of SkillDialog.

### Tests

### Feature Plan

#### General

#### Bots

#### Deployment Scripts

### DIFF
--- a/lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
+++ b/lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
@@ -315,9 +315,6 @@ namespace Microsoft.Bot.Builder.Skills
         {
             try
             {
-                // populate call id for auth purpose
-                activity.CallerId = _serviceClientCredentials.MicrosoftAppId;
-
                 var endOfConversationActivity = await _skillTransport.ForwardToSkillAsync(_skillManifest, _serviceClientCredentials, innerDc.Context, activity, GetTokenRequestCallback(innerDc), GetFallbackCallback(innerDc));
 
                 if (endOfConversationActivity != null)
--- a/lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillWebSocketTransport.cs
+++ b/lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillWebSocketTransport.cs
@@ -54,6 +54,9 @@ namespace Microsoft.Bot.Builder.Skills
 
             await _streamingTransportClient.ConnectAsync(headers);
 
+            // populate call id for auth purpose
+            activity.CallerId = serviceClientCredentials.MicrosoftAppId;
+
             // set recipient to the skill
             var recipientId = activity.Recipient.Id;
             activity.Recipient.Id = skillManifest.MSAappId;


lzc850612:
why do you want to set the callid in the transport class instead of the SkillDialog? if you set it in the transport class, then in the future any newly added transport would have to do this extra duplicate task. SkillDialog is the single entry into the SkillInvocation so we should set callid in SkillDialog.

lzc850612 (changes requested):
why do you want to set the callid in the transport class instead of the SkillDialog? if you set it in the transport class, then in the future any newly added transport would have to do this extra duplicate task. SkillDialog is the single entry into the SkillInvocation so we should set callid in SkillDialog.

lzc850612:
Just realize that you're right since we have CancelRemoteDialogsAsync calls that are not called by ForwardToSkill calls we need this at a lower level.

lzc850612 (approved):

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
```diff
@@ -315,8 +315,6 @@ private async Task<DialogTurnResult> ForwardToSkillAsync(DialogContext innerDc,
         {
             try
             {
-                // populate call id for auth purpose
-                activity.CallerId = _serviceClientCredentials.MicrosoftAppId;
```
lzc850612: remove the empty line

xieofxie: Removed. I wrote too less in issue..
Just be curious about the purpose of CallerId. It is added too months ago without any usage. And will this modification violates the description *'This field is not intended to be transmitted over the wire'*?

lzc850612: The caller id is added in the SDK specifically for the skill invocation scenario. Its purpose is to add additional layer of security because the Skill will perform validation between this call id and the one that's included in the JWT token claims. With this additional validation it's harder for malicious attacker to impersonate calling bot as you may be able to tamper with any property in the Activity object, but you can't generate a JWT token with another app id unless you have its password.

lzc850612: The call id is added specifically for the Skill invocation scenario. Its purpose is to enable additional layer of security. By adding call id in the activity, the skill can validate whether it equals to the appid claim of the JWT token. This way it's a lot harder for malicious caller to impersonate the legitimate caller because it's impossible to get a JWT token with an app id that's not your app unless you have a password.

ryanisgrig (approved):

lib/csharp/microsoft.bot.builder.skills/Microsoft.Bot.Builder.Skills/SkillDialog.cs
xieofxie: Got it, thanks~

PR #2313, opened 2019-09-06
Latest Android Updates

### LINKED ISSUE #2198: [Android] Conversation history is lost when switching to another app
When using the Android VA app - if the active app changes away (e.g. when Google Maps is launched at the end of the PoI skill) and you switch back to the VA app then the conversation history and context is all lost.

The App should maintain conversation history on the canvas even if the app loses focus or the device goes into standby. If you close the App then it's fine to lose the history.

### LINKED ISSUE #2251: [Android] Conversation history is lost when screen being rotated
Conversation history should be kept when the screen being rotated from landscape to portrait or vice versa.

### LINKED ISSUE #2261: Adaptive Cards integration
- [x] A user should be able to interact with the clickable containers/columns on Adaptive Cards
**Currently only a root-level select action is enabled**
- [x] Default card appearance should be transparent and use the background of the bot bubble color
*Right now it is always white, no matter the `app:cardBackgroundColor` set on item_adaptive_card.xml*
- [x] Replace the `hostconfig.json` with empty {} to use default values
- [x] Investigate and potentially replace the HorizontalScrollView/LinearLayout implementation that allows for carousel Adaptive Cards with a RecyclerView (https://www.android-examples.com/android-horizontal-recyclerview-cardview/), not sure if this is causing an issue with the current rendering.
- [x] Adaptive Cards should be rendered in separate bubbles from their related text. This will let us scale the card view independent of related messages. See the BF Emulator for an example of how this looks
image

chon219:
### Purpose
Close #2198: Fixed conversation history lost when switching to other apps (e.g. navigation app, music app)
Close #2251: Fixed conversation history lost when the screen being rotated (landscape to portrait or vice versa)
Close #2261: Fixed several adaptive cards rendering issues
Added newly designed app launch icon as well

### Changes
N/A

### Tests
Not introduced in this project.

### Feature Plan
N/A

#### General


samples/clients/android/VirtualAssistantClient/app/build.gradle
```diff
@@ -42,7 +42,6 @@ dependencies {
     implementation 'org.greenrobot:eventbus:3.1.1'
 
     // for adaptive card rendering, see https://mvnrepository.com/artifact/io.adaptivecards/adaptivecards-android
-    implementation 'io.adaptivecards:adaptivecards-android:1.2.1'
```
ryanisgrig, line 45: Thinking about it, I think we should revert back to using the published library rather than using our own hotfix. We can file the bug in Adaptive Cards repo and wait for the official solution.

samples/clients/android/VirtualAssistantClient/directlinespeech/build.gradle
```diff
@@ -57,4 +57,5 @@ dependencies {
     implementation 'com.jakewharton.rxbinding2:rxbinding:2.0.0'
     implementation 'io.reactivex.rxjava2:rxjava:2.1.9'
     implementation 'io.reactivex.rxjava2:rxandroid:2.0.1'
+    implementation 'com.microsoft.cognitiveservices.speech:client-sdk:1.6.0'
```
ryanisgrig, line 60: What do we need this for that didn't work before?

ryanisgrig (changes requested):

samples/clients/android/VirtualAssistantClient/app/build.gradle
chon219: sure, you are right.

samples/clients/android/VirtualAssistantClient/directlinespeech/build.gradle
chon219: our previous one is a .aar library, so I think we should use the official one instead.

samples/clients/android/VirtualAssistantClient/app/build.gradle
chon219: Reverted back

ryanisgrig (approved):

lauren-mills (approved):

PR #2452, opened 2019-09-28
Removed duplicated MainResponse

tiagodenoronha:


lauren-mills (approved):

lauren-mills (changes requested):
Can you please also make this change in the virtual assistant sample? We will regenerate the template from the sample before the next release

tiagodenoronha:
@lauren-mills done! :)

lauren-mills (approved):

Issue #2639, opened 2019-11-02
Support Local Debugging without Authorization or Secrets


ericmaino:
#### What project is affected?
Microsoft.Bot.Builder.Skills

#### What language is this in?
C#

#### What happens?
Recent changes that added IWhitelistAuthenticationProvider to SkillController and SkillWebSocketAdapter prevents disconnected local debugging.

#### What are the steps to reproduce this issue?
Attempt to run SkillsController or SkillWebSocketAdapter locally without specifying MicrosoftAppId and MicrosoftAppPassword

#### What were you expecting to happen?
Code should support local debugging without network. My current bot project supported this until updating to the most recent version.

#### Any screenshots or additional context?
Suggested changes
- Abstract the code which establishes MsJWTAuthenticationProvider and Authenticator inro a common library, to avoid duplication and to enable dependency injection
- Update SkillController and SkillWebSocketAdapter to take IAuthenicator instead of IWhitelistAuthenticationProvider.

sharath-collab:
Hi,

Even we are facing the same issue when try to debug the latest skill sample project with out specifying app id & app password. Facing below issue. 

Ideally before creating azure resources, we have to be sure it works using bot emulator.
In earlier versions we used to before test it by leaving the app id & app secret blank. Kindly help here

ArgumentNullException: Value cannot be null.
 Parameter name: microsoftAppId

Microsoft.Bot.Builder.Skills.Auth.MsJWTAuthenticationProvider..ctor(string microsoftAppId, string openIdMetadataUrl) in MsJWTAuthenticationProvider.cs, line 20

lzc850612:
So when you are trying to enable local debugging of skill, can you tell me how you configure the VA side to work locally with the skill? The reason I'm asking is because in order to talk to a skill, with the current implementation, you have to specify a skill's app id. otherwise the skill dialog won't work. only when i understand your full scenario then i can have a solution for you.

Thanks.

ericmaino:
@lzc850612 added context PR #2679

PR #2679, opened 2019-11-09
Support Local Debugging without Authorization or Secrets

ericmaino:
### Purpose
Support for local debugging without auth or secrets #2639 

### Changes
Addition of interface and separation of controller to support DI

### Tests
Covered by existing tests. Primarily refactoring.

#### General

#### Bots
- [] I have validated that new and updated responses use appropriate [Speak](https://docs.microsoft.com/en-us/azure/bot-service/dotnet/bot-builder-dotnet-text-to-speech?view=azure-bot-service-3.0) and [InputHint](https://docs.microsoft.com/en-us/azure/bot-service/dotnet/bot-builder-dotnet-add-input-hints?view=azure-bot-service-3.0) properties to ensure a high-quality speech-first experience

#### Deployment Scripts


lzc850612:
So when you are trying to enable local debugging of skill, can you tell me how you configure the VA side to work locally with the skill? The reason I'm asking is because in order to talk to a skill, with the current implementation, you have to specify a skill's app id. otherwise the skill dialog won't work. only when i understand your full scenario then i can have a solution for you.

Thanks.

ericmaino:
All of the skills local manifest in my project look like this which enables easy debugging of the VA and any 1 skill. If I wish to test multiple skills then I modify the port numbers only.

`
    {
      "id": "Calendar",
      "msaAppId": "00000000-0000-0000-0000-000000000000",
      "name": "Calendar Skill",
      "endpoint": "http://localhost:60001/api/skill/messages"
    }
`

The current implementation in the open PR enables me in a local debugging scenario to supply an Anonymous IAuthenticator which works well. I just don't like the idea of shadowing platform types as it leads to a lot of mistakes

lzc850612:
> All of the skills local manifest in my project look like this which enables easy debugging of the VA and any 1 skill. If I wish to test multiple skills then I modify the port numbers only.
> 
> `{ "id": "Calendar", "msaAppId": "00000000-0000-0000-0000-000000000000", "name": "Calendar Skill", "endpoint": "http://localhost:60001/api/skill/messages" }`
> 
> The current implementation in the open PR enables me in a local debugging scenario to supply an Anonymous IAuthenticator which works well. I just don't like the idea of shadowing platform types as it leads to a lot of mistakes

Understood. 

To disable Auth for Skill Invocation flow, there's two sides of the scenario. One is the VA and one is the skill. For VA, the question is whether or not it goes and retrieves a token before calling the skill. For skill, it's a matter of turning on or off the authentication. It looks like right now you only care about disabling auth on the skill side. To achieve that and to be aligned with SDK pattern, the right thing to do should be to use the existing ICredentialProvider and override its IsAuthenticationDisabledAsync function to indicate whether or not skill needs to authenticate incoming skill call. Your PR looks good but it's not implemented this way. Let me make the change and see if it can get into the release we have the coming Monday.

Thanks.

lzc850612:
I've created PR: #2692 to solve the problem of enabling local debugging of skill. Please take a look

Issue #2432, opened 2019-09-25
[Skill] Add spell check feature to LUIS call


bobokids:
## User Story
As ...bot user
I want ...to enable spell check when calling LUIS api 
so that ...typo can be auto corrected in the LUIS call

## Acceptance Criteria 
*List your requirements.*
- [ ] Enable the spell check feature in LUIS call
- [ ] ...
- [ ] ...

## Resources
*Provide any screenshots or conversation mockups to support your request.*
*Learn how to mock up bot conversations with [.chat files](http://aka.ms/bfchatfile).*

## Notes
*Provide any additional information.*

bobokids:
User may refer to implementation in calendar skill to add the spell check support.

PR #2687, opened 2019-11-12
[Skills] Add spell check for skills

DingmaomaoBJTU:
### Purpose
Add spell check for skills
Related #2432

### Changes

### Tests

### Feature Plan

#### General

#### Bots

#### Deployment Scripts


skills/csharp/emailskill/Dialogs/EmailSkillDialogBase.cs
```diff
@@ -1292,6 +1292,8 @@ protected async Task DigestEmailLuisResult(DialogContext dc, bool isBeginDialog)
                 var entity = luisResult.Entities;
                 var generalEntity = generalResult.Entities;
 
+                string luisResultText = string.IsNullOrEmpty(luisResult.AlteredText) ? luisResult.Text : luisResult.AlteredText;
```
lauren-mills, line 1295: I have concerns with using the spellcheck result in all scenarios. We've tested this feature in the past and found that sometimes it corrects words incorrectly, which can cause issues for the user. For instance, if the user wants to invite someone to a meeting with a unique name or email address, but spellcheck corrects it, then the contacts API can't find it, there's no way for the user to correct the issue if we always default to the spellcheck result. I think we need to think more about the design here.

lauren-mills (changes requested):

skills/csharp/emailskill/Dialogs/EmailSkillDialogBase.cs
DingmaomaoBJTU: @lauren-mills yes we notice that too. This will impact user name and email address entities mostly which means skills other than calendar and email are OK. But since this feature can be enable/disabled by skill users in setting, how about we add this support in code(as this is a known luis new feature) and talk to luis team about this issue too?

lauren-mills: @darrenj - thoughts on this?

darrenj: Agreed - we shouldn't default to this as it causes more issues than it solves for the most part. If we were to add this I would want us to test this more significantly in our test environment and work with LUIS on changes that are required.

bobokids:
Close this pr for more investigation. User can refer to calendar skill if he/she wants to add the spell check feature.

PR #2688, opened 2019-11-12
Update onboarding event check to fix null referemce exception

garypretty:
Found I was getting a null reference exception when using some of the new adapters, which was due to an event activity being sent but with a null value.


lauren-mills:
This action behavior has actually been removed in the next branch, so we won't have this problem after release.

Issue #754, opened 2019-02-11
[VA Template] Active Learning


hansmbakker:
**Is your feature request related to a problem? Please describe.**
QnA Maker made [Active Learning](https://docs.microsoft.com/en-us/azure/cognitive-services/qnamaker/how-to/improve-knowledge-base) available. This allows to offer suggested questions to the user if QnA Maker is not certain, and to learn from the question feedback that the user gave.

There is an example implementation available (but without Telemetry integration) at https://github.com/Microsoft/BotBuilder-Samples/blob/master/experimental/csharp_dotnetcore/qnamaker-activelearning-bot/

The Enterprise Bot Template (`TelemetryQnAMaker` class) does not make use of this functionality yet.

**Describe the solution you'd like**
Integration of Active Learning in `TelemetryQnAMaker` class

**Describe alternatives you've considered**
Write it myself.

saurav2107:
Hey team, is there any documentation on how to achieve active learning for QnA in Enterprise Bot template?

lauren-mills:
Hi @saurav2107 we don't have any documentation specific to the enterprise template, but there is a bot sample available here that can be modified to enable the scenario: https://github.com/Microsoft/BotBuilder-Samples/tree/master/experimental/csharp_dotnetcore/qnamaker-activelearning-bot

It looks like this dialog code would do the job: https://github.com/Microsoft/BotBuilder-Samples/blob/master/experimental/csharp_dotnetcore/qnamaker-activelearning-bot/Helpers/DialogHelper.cs

You can replace the qna logic in MainDialog.cs with a BeginDialog() call to this dialog. If you get something working feel free to make a PR!

JackStrap:
Hi @saurav2107 I was able to add active learning for QnA in Enterprise Bot Template.
You can check it at this address : [https://devbolsewebbot.azurewebsites.net/](https://devbolsewebbot.azurewebsites.net/)
To make it work i created a folder "QnA" under Dialogs and i copy, rename the files from the experimental project. The files that we need are under the Helpers folder.
I copy the two FeedbackRecord.cs in my QnA folder
I rename CardHelper.cs to QnACard.cs
I rename DialogHelper.cs to QnADialog.cs
I rename ActiveLearningHelper.cs to QnAResponses.cs
[...]

PR #2809, opened 2019-12-12
Update VA to use the new SDK QnaMakerDialog

### LINKED ISSUE #1888: Support multi-turn QnA Maker in VA
## User Story
As ... a bot developer
I want ... a want to be able to enable multi-turn QnA Maker knowledgebases in VA
so that ... I can enable multi-turn scenarios as needed.

## Acceptance Criteria 
*List your requirements.*
- [ ] Modify existing QnA logic to support either single-turn QnA or multi-turn QnA
- [ ] Should align to SDK implementation
- [ ] Validate with an example in FAQ,lu
- [ ] Should be able to exit "multi-turn" mode at will
- [ ] Should exit multi-turn mode when answer not found

## Resources

## Notes
Multi-turn QnA has the side affect of trying to keep control of the conversation flow after each turn. We need a good way to allow the user to exit multi-turn mode as needed.

Should reference this sample to get started: https://github.com/microsoft/BotBuilder-Samples/tree/master/experimental/qnamaker-prompting/csharp_dotnetcore

darrenj:
# QnAMakerDialog Support

Closes #1888, #754

- Updated Virtual Assistant to use the QnAMakerDialog introduced in SDK R7 which provides Multi-Turn and Active Learning capabilities out of the box. This replaces direct calls to QnAMaker that were provided before
- Added new LG responses for the Active Learning card information to ensure it's localised
- Updates LG response generation in Solutions package to reflect new SDK R7 changes
- Tested with a KB with follow-up prompts
- Obsoleted the `CognitiveModelSet.QnAServices` property as we don't use QnAMaker services, instead we just store the endpoint configuration to enable dialogs to be created. This will nudge people towards a doc item that explains the code change required to an existing VA template.
- Additional docs to guide on the manual changes to existing templates

Note that this will break the Build as it relies on a Solutions lib change. Once PRed we will create a new build and create a new PR referencing the latest bits to avoid the change.


lzc850612 (approved):

docs/_docs/virtual-assistant/handbook/template-updates.md
```diff
@@ -0,0 +1,103 @@
+---
```
lauren-mills: We have a whats new section where we can add these notes and have them easier to find for each release. Might fit better there than in the handbook and can have an associated date.

lauren-mills (changes requested):

docs/_docs/virtual-assistant/handbook/template-updates.md
darrenj: Fair point, I spoke to @ryanlengel yesterday - any thoughts, happy to move around if we think it's a better place?

ryanisgrig: I agree with @lauren-mills, it's better to have a single area of docs that would be shared in chronological order.

ryanisgrig: On the page for this release it can have a section called "Template updates" with this existing doc.

darrenj: Moved.

lauren-mills (approved):
Made some small changes to the metadata of the doc so it will show up in the right place

PR #2851, opened 2019-12-27
Combined VA and Skill templates into one VSIX

lauren-mills:
### Purpose
This change simplifies installation of the VA and Skill templates, and well as reducing maintenance required to manage two installers. 

### Changes
- Combined VA and Skill templates into one
- Moved VA and Skill samples into the samples directory
- Updated templates to include tags for VS2019

### Details
The new VSIX installs both the skill and va templates at the same time (but they are still separate templates)

This is what will show in the new project window when the new vsix is installed. Note, in this case the old skill template is still installed, and two skill templates can be seen:
image

A benefit of the added tags to the templates is that they can be found by selecting the 'Bot' category from the dropdown:
image

The vsix has been renamed, but still has the same id as the existing Virtual Assistant Template, so anyone with the existing template will get the new one when they update and our metrics won't be disrupted:
image


ryanisgrig:
Is the intent behind the VA Sample here to still be a direct copy of a new VA project or can it have customizations as well (i.e. different QnA chitchat personality)?

lauren-mills:
@ryanlengel - still a direct copy for now. I'm planning to use it the same way as before, i.e. exporting from the sample to update the vsix. Just thought it should be in the samples folder for clarity

ryanisgrig (approved):

lzc850612 (approved):

ryanisgrig (changes requested):
I have some changes in a branch that I will PR against so the CI builds work, let's wait for that incase there are other template changes we need to make

ryanisgrig (approved):
@lauren-mills we can merge this one. I will have some feedback based on the nightly builds but my branch against yours had more files changed than it's worth. After this is in I can update the VASample demo builds to reflect the new directory structure and then continue work on the nightly validation.

PR #2860, opened 2020-01-08
[Sdk] use delegate instead of string in FeedbackOptions

### LINKED ISSUE #2516: FeedbackOptions should support multi-language and different FeedbackReceivedMessage
## User Story
As ... a dev
I want ... that FeedbackOptions should support multi-language and different FeedbackReceivedMessage
so that ... users feel better Feedback experience

## Acceptance Criteria 
*List your requirements.*
- [ ] ... feedbackReceivedMessage, commentPrompt, feedbackReceivedMessage should be delegate instead of string, so multi-language could be supported
- [ ] ... feedbackReceivedMessage should return different ones according to feedback like "sorry about that" for thumbs down (if it is a delegate with result as parameter, this is easy)
- [ ] ...

## Resources

## Notes

From CUI Skill Reviews

xieofxie:
### Purpose

Close #2516

### Changes

* Change FeedbackRecord to internal
* Use delegate instead of string in FeedbackOptions
* FeedbackReceivedMessageDelegate returns a bool to indicate whether prompt for comments
(Struggle between generalization and simplicity..)

### Tests

* Add DefaultOptions_CommentDirectly, CustomOptions_Neutral

### Feature Plan

#### General

#### Bots

#### Deployment Scripts

### DIFF
--- a/sdk/csharp/libraries/microsoft.bot.solutions/Feedback/FeedbackConstants.cs
+++ b/sdk/csharp/libraries/microsoft.bot.solutions/Feedback/FeedbackConstants.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Bot.Solutions.Feedback
+{
+    public static class FeedbackConstants
+    {
+        // The name of the event when a feedback is received
+        public static readonly string FeedbackEvent = "Feedback";
+
+        public static readonly string FeedbackTag = "feedbackTag";
+
+        public static readonly string FeedbackValue = "feedbackValue";
+
+        public static readonly string FeedbackComent = "feedbackComment";
+    }
+}
--- a/sdk/csharp/libraries/microsoft.bot.solutions/Feedback/FeedbackMiddleware.cs
+++ b/sdk/csharp/libraries/microsoft.bot.solutions/Feedback/FeedbackMiddleware.cs
@@ -60,6 +60,11 @@ namespace Microsoft.Bot.Solutions.Feedback
             // store in state. No need to save changes, because its handled in IBot
             await _feedbackAccessor.SetAsync(context, record).ConfigureAwait(false);
 
+            var allActions = new List<CardAction>(_options.FeedbackActions(context, tag))
+            {
+                _options.DismissAction(context, tag),
+            };
+
             // If channel supports suggested actions
             if (Channel.SupportsSuggestedActions(context.Activity.ChannelId))
             {
@@ -69,21 +74,21 @@ namespace Microsoft.Bot.Solutions.Feedback
                 {
                     var actions = new List<CardAction>()
                         .Concat(context.Activity.SuggestedActions.Actions)
-                        .Concat(GetFeedbackActions())
+                        .Concat(allActions)
                         .ToList();
 
                     await context.SendActivityAsync(MessageFactory.SuggestedActions(actions)).ConfigureAwait(false);
                 }
                 else
                 {
-                    var actions = GetFeedbackActions();
+                    var actions = allActions;
                     await context.SendActivityAsync(MessageFactory.SuggestedActions(actions)).ConfigureAwait(false);
                 }
             }
             else
             {
                 // else channel doesn't support suggested actions, so use hero card.
-                var hero = new HeroCard(buttons: GetFeedbackActions());
+                var hero = new HeroCard(buttons: allActions);
                 await context.SendActivityAsync(MessageFactory.Attachment(hero.ToAttachment())).ConfigureAwait(false);
             }
         }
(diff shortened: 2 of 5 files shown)


lauren-mills:
@xieofxie - Please address the stylecop errors that are breaking the build. Thanks!

xieofxie:
@lauren-mills oops.. fixed

sdk/csharp/libraries/microsoft.bot.builder.solutions/Feedback/FeedbackOptions.cs
```diff
@@ -24,115 +26,64 @@ public class FeedbackOptions
...
         /// </summary>
         /// <value>
-        /// Message to show when a user provides some feedback.
+        /// Message to show and wether to prompt for comments when a user provides some feedback.
         /// </value>
-        public string FeedbackReceivedMessage
+        public FeedbackReceivedMessageDelegate FeedbackReceivedMessage { get; set; } = (ITurnContext context, string tag, CardAction action) =>
         {
-            get
+            if ((string)action.Value == "negative")
```
darrenj: should be a constant?

sdk/csharp/libraries/microsoft.bot.builder.solutions/Feedback/FeedbackOptions.cs
```diff
@@ -24,115 +26,64 @@ public class FeedbackOptions
...
         /// Text to show on button that allows user to hide/ignore the feedback request.
         /// </value>
-        public CardAction DismissAction
+        public DismissActionDelegate DismissAction { get; set; } = (ITurnContext context, string tag) =>
         {
-            get
-            {
-                if (this.dismissAction == null)
-                {
-                    return new CardAction(ActionTypes.PostBack, title: FeedbackResponses.DismissTitle, value: "dismiss");
```
darrenj: string constants?

sdk/csharp/libraries/microsoft.bot.builder.solutions/Feedback/FeedbackOptions.cs
```diff
@@ -24,115 +26,64 @@ public class FeedbackOptions
...
+        public FeedbackActionsDelegate FeedbackActions { get; set; } = (ITurnContext context, string tag) =>
         {
-            get
-            {
-                if (this.feedbackActions == null)
+            return new List<CardAction>()
                 {
-                    return new List<CardAction>()
-                    {
-                        new CardAction(ActionTypes.PostBack, title: "👍", value: "positive"),
```
darrenj: string constants?

darrenj (changes requested):

xieofxie:
Feeling a sense of over-consideration..
```
FeedbackOptions
* Add LogPersonalInformation <- follow TelemetryLoggerMiddleware
* Add Default*Values
* Change string to IMessageActivity
FeedbackMiddleware
* Save FeedbackMiddleware in TurnState instead of static variables
* Use TelemetryConstants instead of nameof <- should update PowerBI template
* Not send when IMessageActivity is null
```

@lauren-mills Could you give an example about when this will happen? Since context.Activity should be the incoming one. Thanks~ https://github.com/microsoft/botframework-solutions/blob/d4d380f30234e7811e30f7ee2137c6175b482749/sdk/csharp/libraries/microsoft.bot.builder.solutions/Feedback/FeedbackMiddleware.cs#L66

darrenj (approved):

darrenj:
@lauren-mills can you review and approve if your OK with the changes?

darrenj (changes requested):
Can you explain a little more why we serialize the whole middleware into TurnState rather than just data you need across turns?

xieofxie:
TurnState is Dictionary<string, object> and it will only hold a reference to the middleware instance. I think serialization/deserialization will not happen during the lifecycle.

Or we could guide the user to use RegisterClassMiddleware to do this (in case the key is used which is quite impossible) and it will be more complex.

darrenj:
I understand TurnState just wanted to clarify precisely what data you need across the lifetime of a turn and consider persisting just that instead of entire middleware which is confusing?

xieofxie:
OK, I add a private class `RequestTurnState` to store the needed data and a public key `RequestTurnStateKey` to indicate that TurnState is modified and should take care.

darrenj:
Thanks! @pavolum could you perhaps share with @xieofxie the issue you ran into with #2979 and see if we can incorporate a fix as part of this?

pavolum:
> Thanks! @pavolum could you perhaps share with @xieofxie the issue you ran into with #2979 and see if we can incorporate a fix as part of this?

Sure thing. @xieofxie feel free to look at #2979 as it has details on the issue. Essentially the method of triggering the feedback middleware in the bot code (calling FeedbackMiddleware.RequestFeedbackAsync(turnContext, "your-tag")) will no longer work with the new waterfall main dialog implementation in the VA. 

The waterfall main dialog makes its such that there is essentially always a listener for the next activity. So when the bot fires off the feedback prompt and a response is returned, both the bot and the feedback middleware are set to process that feedback response. The bot has no logic to handle this response and it gets into a strange state.

At first glance I think the solution could be to convert the feedback middleware into a dialog that can be called from a VA or skill project. That way the feedback multi turn dialog is a part of the overall bot waterfall dialog life cycle. However there could be a less involved change if we get creative (perhaps keep the middleware but have a dialog wrapper for it, however at that point we may as well make the whole thing a dialog thats part of the original template or the solutions lib). If it is a larger refactor needed I would suggest having a separate PR for it.

If you want further understanding I suggest trying to fire a feedback request from anywhere within a 0.8 VA using the documented method (https://microsoft.github.io/botframework-solutions/virtual-assistant/handbook/feedback/). This should reproduce the issue.

pavolum:
@xieofxie have you started any work to address #2979 which I outlined in a previous comment? If not please hold off. We have decided that we will create a spec for the ideal feedack mechanism in the next few days and then priorities when it will be implemented and by which team. Thanks!

PR #2882, opened 2020-01-15
Doc updates for Manifest and Action enhancements to an existing Skill

darrenj:
- Updates to the migration to BF 4.7 Skills section enabling an existing skill developer to update their manifest and add action handling code


ryanisgrig (approved):
Nit: Replace the numbered list with all 1's and it will auto increment the steps

darrenj:
> Nit: Replace the numbered list with all 1's and it will auto increment the steps

Done!

PR #2917, opened 2020-01-27
Add New Skill Manifest to all Skills

darrenj:
### New Skill Manifest

SDK has introduced a new Skill Manifest as part of their GA of Skills. Updated all Skills with the new manifest alongside the existing manifestTemplate.json file which will be removed in the next release.

No actions are defined as support for these hasn't been implemented in each skill therefore only activity/utterance invocation is supported as per before.

### DIFF
--- a/skills/csharp/Skills.sln
+++ b/skills/csharp/Skills.sln
@@ -1,4 +1,4 @@
-﻿﻿
+﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 16
 VisualStudioVersion = 16.0.29102.190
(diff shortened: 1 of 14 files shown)


skills/csharp/Skills.sln
```diff
@@ -51,7 +51,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ToDoSkill.Tests", "tests\to
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ITSMSkill.Tests", "tests\itsmskill.tests\ITSMSkill.Tests.csproj", "{C3304FB4-D4D5-412F-9312-1ECE4DE573DA}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "HospitalitySkill.Tests", "tests\HospitalitySkill.Tests\HospitalitySkill.Tests.csproj", "{E3E37DC8-1AEC-4368-A7F0-62F93A82FDD0}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "HospitalitySkill.Tests", "tests\HospitalitySkill.Tests\HospitalitySkill.Tests.csproj", "{E3E37DC8-1AEC-4368-A7F0-62F93A82FDD0}"
```
lzc850612, line 54: this doesn't seem necessary. probably auto-generated?

lzc850612 (changes requested):
only comment on .sln file auto-generated change and everything else looks great.

skills/csharp/Skills.sln
darrenj: Indeed - that said it's I think corrected a mismatch between that project and everything else so seems reasonable

Issue #2996, opened 2020-02-06
[POI] add 'best match' to poiType(PoiDescription)


xieofxie:
## User Story
As ... anyone
I want ... to use action without interruption
so that ... it is easy to use in PVA etc.

## Acceptance Criteria 
*List your requirements.*
- [ ] ... add 'best match' to poiType, so the following choosing step is skipped
![image](https://user-images.githubusercontent.com/2876650/73937396-ebf90200-491f-11ea-8746-7b976f3078de.png)
- [ ] ...
- [ ] ...

## Resources
*Provide any screenshots or conversation mockups to support your request.*
*Learn how to mock up bot conversations with [.chat files](http://aka.ms/bfchatfile).*

## Notes
*Provide any additional information.*

Issue #2995, opened 2020-02-06
[POI] add actions


xieofxie:
## User Story
As ... anyone
I want ... to POI supports actions
so that ... it could be used in PVA etc.

## Acceptance Criteria 
*List your requirements.*
- [ ] ... add actions to manifest
- [ ] ... implement actions
- [ ] ... test with PVA, Composer
- [ ] ... add unit tests for actions

## Resources
*Provide any screenshots or conversation mockups to support your request.*
*Learn how to mock up bot conversations with [.chat files](http://aka.ms/bfchatfile).*

## Notes
*Provide any additional information.*

PR #2997, opened 2020-02-06
[POI] add actions to manifest

xieofxie:
### Purpose

First step of #2995

### Changes

* add 3 actions
* add inputObject, responseObject
    - showRoute is used to skip the following choosing step, so action could execute to the end without interruption
image

### Tests

### Feature Plan

Related #2996

#### General

#### Bots

#### Deployment Scripts


skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
```diff
@@ -23,9 +23,98 @@
...
+        "$ref": "#/definitions/responseObject"
+      }
+    },
     "message": {
       "type": "message",
       "description": "Receives the users utterance and attempts to resolve it using the skill's LU models"
     }
+  },
+  "definitions": {
+    "inputObject": {
```
darrenj: we don't need to call them input and outputObject if you'd prefer another name :)

skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
```diff
@@ -23,9 +23,98 @@
...
+          "type": "boolean",
+          "description": "Whether show route."
+        }
+      }
+    },
+    "responseObject": {
+      "type": "object",
+      "properties": {
+        "latitude": {
+          "type": "number",
```
darrenj, line 172: Could we also return poi name too?

darrenj: I think responseObject will have to be an array? i.e. multiple items?

darrenj (changes requested):

skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
xieofxie: Since all three actions use same input/output, I just keep them.

skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
xieofxie: Add name and address. I just follow the current dialog and it will only return one in the end.
Should we add a flag to break the flow or change it to array for broader compatibility?

darrenj: Ahh OK - This is to to find items, let the user choose on and then return one item? Makes sense. Thanks!

skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
```diff
@@ -23,9 +23,106 @@
...
+        "$ref": "#/definitions/inputObject"
+      },
+      "resultValue": {
+        "$ref": "#/definitions/responseObject"
+      }
+    },
+    "findParkingAction": {
+      "description": "Find parking, category will be ignored.",
+      "type": "event",
+      "name": "FindParkingAction",
```
darrenj, line 51: In this case we should define a different object type with just the required "Input data" items. This way things like composer only prompt for the items the operation will actually need, does that make sense?

skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
darrenj: Good point - let's leave it as it is now. It might be helpful to have a "FindPoI" action that returns options to the caller which can do it's own logic and not use the rest of the skill capabilities.

darrenj (changes requested):
minor change below otherwise looks good! I've changed the PR to point to next :)

skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
lzc850612: Agreed. Do all three dialogs always require the same set of parameters? if not we should define separate inputObject for them.

Should we also add this in 1.1 manifest so VAs can also use it?

lzc850612 (changes requested):

skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
xieofxie: OK, change to 3 different inputs. It seems [allOf](https://json-schema.org/understanding-json-schema/structuring.html) is not supported, so similar parts should be maintained manually.

skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
```diff
@@ -23,9 +23,168 @@
     }
   ],
   "activities": {
+    "getDirectionAction": {
+      "description": "Get destination of some poi.",
```
darrenj: "Get directions to a provided Point Of Interest"

skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
darrenj: Thanks. Can we also get this replicated into the mainifest-1.1.json manifest too (1.0 is just for P-VA). Thanks!

darrenj (changes requested):

skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
xieofxie: Done.

skills/csharp/pointofinterestskill/wwwroot/manifest/manifest-1.0.json
xieofxie: Done.

darrenj (approved):
thanks! look forward to the action support being added now :-)

xieofxie:
@lzc850612 please check this, thanks~

bobokids (approved):

lzc850612 (approved):

PR #3020, opened 2020-02-10
[POI] filter entity type by favoredEntity

### LINKED ISSUE #3008: [PoI] Multiple responses incorrectly provided for location selection
# Issue

The [GetPointOfInterestListByAddressAsync](https://github.com/microsoft/botframework-solutions/blob/master/skills/csharp/shared/SkillServiceLibrary/Services/AzureMapsAPI/AzureMapsGeoSpatialService.cs#L175) method handles talking a location from the user (e.g. Bellevue, WA, London or London, UK) and trying to resolve a location to then do a POI search.

For example: 

image

image

# Potential resolution

For Ignite (within a Composer variant of PoI) I experimented with this change which ensures we only focus on municipalities rather than "geography areas" and other more broad results which come back from the [Azure Maps API](https://docs.microsoft.com/en-us/rest/api/maps/search/getsearchaddress). 

The [Azure Maps API](https://docs.microsoft.com/en-us/rest/api/maps/search/getsearchaddress) doesn't seem to provide any API level control over the types of entities returned.

In this example I then only picked the first one which is likely not the right logic for a formal fix but may help with looking at options:

```csharp
public async Task<IEnumerable<SearchResult>> GetLocation(string location)
        {
            string poiCategoryUri = string.Format(FindByAddressNoCoordinatesQueryUrl, location, limit);
            poiCategoryUri += $"&language={locale}&subscription-key={apiKey}";

            var response = await httpClient.GetStringAsync(poiCategoryUri);
            var apiResponse = JsonConvert.DeserializeObject<SearchResultSet>(response);

            if (apiResponse?.Results != null)
            {
                // We only care about Municipalities, not sub-regions. Distinct important 
                var locations = apiResponse.Results
                     .Where(e => e.EntityType == "Municipality")
                     .GroupBy(e => e.Address.FreeformAddress)
                     .Select(a => a.FirstOrDefault());

                return locations;
            }
            else
            {
                return null;
            }
        }
```

# Success Criteria

I can ask the following questions and *not* be prompted to choose from multiple locations I think we are in good shape.

- Find a coffee shop in London
- Find a Pharmacy in bellevue
- Find a walgreen store in Seattle

xieofxie:
### Purpose

Close #3008

### Changes

* filter entity type by Municipality when searching for current location
    - if no Municipality, fallback to all
* add EntityType
* separate tests

### Tests

### Feature Plan

#### General

#### Bots

#### Deployment Scripts


skills/csharp/pointofinterestskill/Dialogs/PointOfInterestDialogBase.cs
```diff
@@ -608,17 +608,6 @@ protected PromptOptions GetPointOfInterestPrompt(string prompt, Card containerCa
             return options;
         }
 
-        // Validators
-        protected async Task<List<PointOfInterestModel>> CurrentLocationValidator(PromptValidatorContext<string> promptContext, CancellationToken cancellationToken)
```
lzc850612, line 612: Some changes in this PR also exist in another PR that's currently open.

lzc850612 (changes requested):

skills/csharp/pointofinterestskill/Dialogs/PointOfInterestDialogBase.cs
xieofxie: Oh, I created two prs to solve two issues separately and #3026 depends on this. Or just view #3026 for two issues together.

darrenj (approved):

xieofxie:
Merged by #3026

PR #3154, opened 2020-03-09
[Skill] handle cancel/logout in skill mode

### LINKED ISSUE #3084: [Skills] handle cancel etc. in skill mode
## User Story
As ... a PVA + skill user
I want ... to cancel to be properly handled
so that ... it is good experience

## Acceptance Criteria 
*List your requirements.*
- [ ] ... Handle cancel etc. properly in skill mode

## Resources

When in PVA a user sends cancel, it will be sent to skill and then the skill begin main dialog again which will parse cancel in RouteStepAsync.
This is not desired logic. We should not begin main dialog in skill mode.

## Notes

xieofxie:
### Purpose

Close #3084

### Changes

* Send EoC with EndOfConversationCodes.UserCancelled in Cancel & Logout in skill mode

### Tests

image

### Feature Plan

#### General

### DIFF
--- a/samples/csharp/skill/SkillSample/Dialogs/MainDialog.cs
+++ b/samples/csharp/skill/SkillSample/Dialogs/MainDialog.cs
@@ -78,10 +78,10 @@ namespace SkillSample.Dialogs
                 // Check for any interruptions
                 var interrupted = await InterruptDialogAsync(innerDc, cancellationToken);
 
-                if (interrupted)
+                if (interrupted != null)
                 {
                     // If dialog was interrupted, return EndOfTurn
-                    return EndOfTurn;
+                    return interrupted;
                 }
             }
 
@@ -109,10 +109,10 @@ namespace SkillSample.Dialogs
                 // Check for any interruptions
                 var interrupted = await InterruptDialogAsync(innerDc, cancellationToken);
 
-                if (interrupted)
+                if (interrupted != null)
                 {
                     // If dialog was interrupted, return EndOfTurn
-                    return EndOfTurn;
+                    return interrupted;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


samples/csharp/skill/SkillSample/Dialogs/MainDialog.cs
```diff
@@ -139,7 +139,21 @@ protected async Task<bool> InterruptDialogAsync(DialogContext innerDc, Cancellat
                             {
                                 await innerDc.Context.SendActivityAsync(_templateEngine.GenerateActivityForLocale("CancelledMessage"));
                                 await innerDc.CancelAllDialogsAsync();
-                                await innerDc.BeginDialogAsync(InitialDialogId);
+                                if (innerDc.Context.IsSkill())
+                                {
+                                    // EndOfConversation activity should be passed back to indicate that VA should resume control of the conversation
+                                    var endOfConversation = new Activity(ActivityTypes.EndOfConversation)
```
darrenj: @lzc850612 I think this is handled by the latest SDK bits?

darrenj (changes requested):

samples/csharp/skill/SkillSample/Dialogs/MainDialog.cs
lzc850612: actually no. SDK doesn't handle interruption.

i think this is worth adding. If people are using our VA then they probably don't need it. But from skill's point of view, skill should make sure to send out EndOfConversation in skill mode when the ongoing dialog is being cancelled, just in case parent bot didn't handle interruption properly.

lzc850612 (approved):

samples/csharp/skill/SkillSample/Dialogs/MainDialog.cs
xieofxie: @darrenj @lzc850612 After looking into the new SDK and consider we need to return something in action mode, I updated the code to use DialogTurnResult directly.

DialogExtensions will send EoC when it ends here:
https://github.com/microsoft/botbuilder-dotnet/blob/373096310de1272ac1f26f2045448b3f40f4ff68/libraries/Microsoft.Bot.Builder.Dialogs/DialogExtensions.cs#L78 

To test it against VA, VA needs to be updated to 4.8.0-rc5 as it fixed a bug

If feel good, I will mirror the updates to template. Thanks

lzc850612: looks good

xieofxie: Mirrored

Issue #3170, opened 2020-03-11
[SDK] incorrect test case GetResponseWithLanguageGeneration_Spanish?


xieofxie:
#### What project is affected?
Solutions SDK
#### What language is this in?
C#
#### What happens?
In this line https://github.com/microsoft/botframework-solutions/blob/6c45f643718df6b2b52593e1792ab50f7c8bad0e/sdk/csharp/tests/microsoft.bot.solutions.tests/Responses/LanguageGenerationTests.cs#L64

It should be es-es instead of en-us? Since we have es-es in Setup
#### Any screenshots or additional context?
This line should be fixed https://github.com/microsoft/botframework-solutions/blob/6c45f643718df6b2b52593e1792ab50f7c8bad0e/sdk/csharp/libraries/microsoft.bot.solutions/Responses/LocaleTemplateManager.cs#L50

But even if we switch order, it is not a real fallback as we don't check if it is in the keys of localeTemplateFiles. Maybe we should wait until bot sdk's fallback is published

lzc850612:
I've got PR fixing the test

for the second point, the Generate function internally maintains all combination of fallback chains so we don't have to check. Guarding logic is in the SDK level.

PR #3172, opened 2020-03-11
fix spanish test

lzc850612:
### Purpose

Fix spanish language generation test

### Changes

### Tests

### Feature Plan

#### General

### DIFF
--- a/sdk/csharp/libraries/microsoft.bot.solutions/Responses/LocaleTemplateManager.cs
+++ b/sdk/csharp/libraries/microsoft.bot.solutions/Responses/LocaleTemplateManager.cs
@@ -26,6 +26,12 @@ namespace Microsoft.Bot.Solutions.Responses
         public LocaleTemplateManager(Dictionary<string, string> localeTemplateFiles, string fallbackLocale)
             : base(localeTemplateFiles)
         {
+            // only throw when fallbackLocale is empty string
+            if (fallbackLocale != null && (fallbackLocale == string.Empty || fallbackLocale.Trim() == string.Empty))
+            {
+                throw new ArgumentException($"{nameof(fallbackLocale)} shouldn't be empty string. If you don't want to set it, please set it to null.");
+            }
+
             _fallbackLocale = fallbackLocale;
         }
 
@@ -45,6 +51,12 @@ namespace Microsoft.Bot.Solutions.Responses
             if (templateName == null)
             {
                 throw new ArgumentNullException(nameof(templateName));
+            }
+
+            // only throw when localeOverride is empty string
+            if (localeOverride != null && (localeOverride == string.Empty || localeOverride.Trim() == string.Empty))
+            {
+                throw new ArgumentException($"{nameof(localeOverride)} shouldn't be empty string. If you don't want to set it, please set it to null.");
             }
 
             var locale = localeOverride ?? _fallbackLocale ?? CultureInfo.CurrentUICulture.Name;
--- a/sdk/csharp/libraries/microsoft.bot.solutions/Responses/LocaleTemplateManager.cs
+++ b/sdk/csharp/libraries/microsoft.bot.solutions/Responses/LocaleTemplateManager.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Bot.Solutions.Responses
         /// <param name="localeTemplateFiles">A dictionary of locale and template file.</param>
         /// <param name="fallbackLocale">The default fallback locale to use.</param>
         public LocaleTemplateManager(Dictionary<string, string> localeTemplateFiles, string fallbackLocale)
-            : base(localeTemplateFiles)
+            : base(fallbackLocale == null ? localeTemplateFiles : new Dictionary<string, string>(localeTemplateFiles) { { string.Empty, localeTemplateFiles[fallbackLocale] } })
         {
             // only throw when fallbackLocale is empty string
             if (fallbackLocale != null && (fallbackLocale == string.Empty || fallbackLocale.Trim() == string.Empty))
@@ -59,7 +59,7 @@ namespace Microsoft.Bot.Solutions.Responses
                 throw new ArgumentException($"{nameof(localeOverride)} shouldn't be empty string. If you don't want to set it, please set it to null.");
             }
 
-            var locale = localeOverride ?? _fallbackLocale ?? CultureInfo.CurrentUICulture.Name;
+            var locale = localeOverride ?? CultureInfo.CurrentUICulture.Name ?? _fallbackLocale;
 
             return ActivityFactory.FromObject(Generate($"${{{templateName}()}}", data, locale));
         }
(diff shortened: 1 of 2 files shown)


lzc850612:
fixes #3170

xieofxie:
Hi @lzc850612 I found that in MultiLanguageLG, it follows the following fallback rule: 
`en-us -> en -> ""`
And together with SetLocaleMiddleware, we should map the lg like
* string.Empty -> default file like en-us.lg
* en -> en-us.lg
* es -> es-es.lg etc.

Then when en-us, en-gb etc. come, they will fallback to en-us.lg, when none of these, it will fallback to empty, thus en-us.lg

A suggestion here: https://github.com/microsoft/botframework-solutions/commit/4dbbe81e8523299f197d8c0d91aa1a4a72a21e75

lzc850612:
> Hi @lzc850612 ted Li FTE I found that in MultiLanguageLG, it follows the following fallback rule:
> `en-us -> en -> ""`
> And together with SetLocaleMiddleware, we should map the lg like
> 
> * string.Empty -> default file like en-us.lg
> * en -> en-us.lg
> * es -> es-es.lg etc.
> 
> Then when en-us, en-gb etc. come, they will fallback to en-us.lg, when none of these, it will fallback to empty, thus en-us.lg
> 
> A suggestion here: [4dbbe81](https://github.com/microsoft/botframework-solutions/commit/4dbbe81e8523299f197d8c0d91aa1a4a72a21e75)

LocaleTemplateManager needs to be able to pass in a default locale, so bot developer can pass in a default locale in startup for its VA. that way the bot developer doesn't have to pass in the locale in every call.

xieofxie:
> > Hi @lzc850612 ted Li FTE I found that in MultiLanguageLG, it follows the following fallback rule:
> > `en-us -> en -> ""`
> > And together with SetLocaleMiddleware, we should map the lg like
> > 
> > * string.Empty -> default file like en-us.lg
> > * en -> en-us.lg
> > * es -> es-es.lg etc.
> > 
> > Then when en-us, en-gb etc. come, they will fallback to en-us.lg, when none of these, it will fallback to empty, thus en-us.lg
> > A suggestion here: [4dbbe81](https://github.com/microsoft/botframework-solutions/commit/4dbbe81e8523299f197d8c0d91aa1a4a72a21e75)
> 
> LocaleTemplateManager needs to be able to pass in a default locale, so bot developer can pass in a default locale in startup for its VA. that way the bot developer doesn't have to pass in the locale in every call.

I assume SetLocaleMiddleware is used to change CultureInfo.CurrentUICulture.Name here.
And to handle non-exist locale in the map, we still need a string.Empty key-value pair.

https://github.com/microsoft/botframework-solutions/commit/54f26287435f55d8fa018bb9f4c96aa7be4bf04d

lzc850612:
> > > Hi @lzc850612
> > >             
> > >             ted Li
> > >             FTE ted Li FTE I found that in MultiLanguageLG, it follows the following fallback rule:
> > > `en-us -> en -> ""`
> > > And together with SetLocaleMiddleware, we should map the lg like
> > > 
> > > * string.Empty -> default file like en-us.lg
> > > * en -> en-us.lg
> > > * es -> es-es.lg etc.
> > > 
> > > Then when en-us, en-gb etc. come, they will fallback to en-us.lg, when none of these, it will fallback to empty, thus en-us.lg
> > > A suggestion here: [4dbbe81](https://github.com/microsoft/botframework-solutions/commit/4dbbe81e8523299f197d8c0d91aa1a4a72a21e75)
> > 
> > 
> > LocaleTemplateManager needs to be able to pass in a default locale, so bot developer can pass in a default locale in startup for its VA. that way the bot developer doesn't have to pass in the locale in every call.
> 
> I assume SetLocaleMiddleware is used to change CultureInfo.CurrentUICulture.Name here.
> And to handle non-exist locale in the map, we still need a string.Empty key-value pair.
> 
> [54f2628](https://github.com/microsoft/botframework-solutions/commit/54f26287435f55d8fa018bb9f4c96aa7be4bf04d)

SetLocaleMiddleware is something that bot developer can use, but they can also not use it. We can't have LocaleTemplateManager depend on that. We need to make sure it works on its own when bot developer only uses it without SetLocaleMiddleware

For empty locale, the only thing we should do is to barf when an empty string is being passed in, because we shouldn't treat string.Empty as 'en-us'. for a component that other people use, we need to be specific and throw an exception when needed instead of doing a fallback that's unknown to people who use it

xieofxie:
But to use the call without explicit set localeOverride, CultureInfo.CurrentUICulture.Name still should be set somewhere.

And the order of `var locale = localeOverride ?? _fallbackLocale ?? CultureInfo.CurrentUICulture.Name;` should be changed because MultiLanguageLG.Generate doesn't know the current locale. It will always be _fallbackLocale in this case.

lzc850612:
> But to use the call without explicit set localeOverride, CultureInfo.CurrentUICulture.Name still should be set somewhere.

Cultureinfo.CurrentUICulture.Name has a default value we don't have to set it. It's the server's default culture

> 
> And the order of `var locale = localeOverride ?? _fallbackLocale ?? CultureInfo.CurrentUICulture.Name;` should be changed because MultiLanguageLG.Generate doesn't know the current locale. It will always be _fallbackLocale in this case.

why would it always be _fallbackLocale? why wouldn't it be localeOverride when calling this function with localeOverride set to be a certain value?

xieofxie:
> > But to use the call without explicit set localeOverride, CultureInfo.CurrentUICulture.Name still should be set somewhere.
> 
> Cultureinfo.CurrentUICulture.Name has a default value we don't have to set it. It's the server's default culture
> 
> > And the order of `var locale = localeOverride ?? _fallbackLocale ?? CultureInfo.CurrentUICulture.Name;` should be changed because MultiLanguageLG.Generate doesn't know the current locale. It will always be _fallbackLocale in this case.
> 
> why would it always be _fallbackLocale? why wouldn't it be localeOverride when calling this function with localeOverride set to be a certain value?

Then the user will always have to fill localeOverride parameter with current locale when calling the function. If the order is switched with CultureInfo.CurrentUICulture.Name set, that parameter doesn't need to be set.

sdk/csharp/libraries/microsoft.bot.solutions/Responses/LocaleTemplateManager.cs
```diff
@@ -24,8 +24,14 @@ public class LocaleTemplateManager : MultiLanguageLG
         /// <param name="localeTemplateFiles">A dictionary of locale and template file.</param>
         /// <param name="fallbackLocale">The default fallback locale to use.</param>
         public LocaleTemplateManager(Dictionary<string, string> localeTemplateFiles, string fallbackLocale)
-            : base(localeTemplateFiles)
+            : base(fallbackLocale == null ? localeTemplateFiles : new Dictionary<string, string>(localeTemplateFiles) { { string.Empty, localeTemplateFiles[fallbackLocale] } })
```
xieofxie, line 27: This line is too complex. Should MultiLanguageLG be changed to a member instead of base class?

xieofxie:
OK, pushed

sdk/csharp/libraries/microsoft.bot.solutions/Responses/LocaleTemplateManager.cs
lzc850612: should be better to keep the inheritance hierarchy. 

let's leave it as-is

xieofxie (approved):

xieofxie:
@lzc850612 one more question: due to the fallback rule, should we update templates and skills to en -> en-us.lg instead of en-us -> en-us.lg?

PR #3218, opened 2020-03-19
SkillSample updates

gabog:
This PR contains several suggested changes to the SkillSample based on the SDK samples and other observations: 

- Refactored DefaultAdapter based on SDK Skill samples (broke down OnTurnError in several discrete methods with error handling and deleted conversation state at the end).
- Added AllowedCallersClaimsValidator sample.
- Moved assignment of MainDialog.TelemetryClient into DefaultActivityHandler to avoid calling a virtual property from the Dialog's constructor.
- Made private fields readonly where possible.
- Removed unused _stateAccessor from MainDialog
- Added missing cancellationToken parameters across the board.
- Added Async suffic to async methods.
- Simplified SDK references by removing redundant packages already included in top packages.
- Made private class DialogIds static
- Updated reference to StyleCop.Analyzers to released version and add the  PrivateAssets="all" property to avoid propagating it to child projects.
- Added a reference toMicrosoft.AspNetCore.Mvc.NewtonsoftJson to the project and updated startup.cs to .AddNewtonsoftJson() to ensure compatibility.
- Updated startup.cs to register the AllowedCallersClaimsValidator.

### DIFF
--- a/samples/csharp/skill/SkillSample/Adapters/DefaultAdapter.cs
+++ b/samples/csharp/skill/SkillSample/Adapters/DefaultAdapter.cs
@@ -1,14 +1,18 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+using System.Threading.Tasks;
 using Microsoft.Bot.Builder;
 using Microsoft.Bot.Builder.Azure;
 using Microsoft.Bot.Builder.Integration.ApplicationInsights.Core;
 using Microsoft.Bot.Builder.Integration.AspNet.Core;
+using Microsoft.Bot.Builder.TraceExtensions;
 using Microsoft.Bot.Connector.Authentication;
 using Microsoft.Bot.Schema;
 using Microsoft.Bot.Solutions.Middleware;
 using Microsoft.Bot.Solutions.Responses;
+using Microsoft.Extensions.Logging;
 using SkillSample.Extensions;
 using SkillSample.Services;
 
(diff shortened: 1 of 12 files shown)


lauren-mills (approved):

samples/csharp/skill/SkillSample/Adapters/DefaultAdapter.cs
```diff
@@ -1,46 +1,48 @@
...
+        private readonly ConversationState _conversationState;
+        private readonly ILogger _logger;
+        private readonly IBotTelemetryClient _telemetryClient;
+        private readonly LocaleTemplateManager _templateEngine;
+
         public DefaultAdapter(
             BotSettings settings,
             IChannelProvider channelProvider,
             ICredentialProvider credentialProvider,
+            AuthenticationConfiguration authConfig,
```
lzc850612, line 32: what do we need AuthenticationConfiguration for?

samples/csharp/skill/SkillSample/Startup.cs
```diff
@@ -137,6 +141,9 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
                 .UseWebSockets()
                 .UseRouting()
                 .UseEndpoints(endpoints => endpoints.MapControllers());
+
+            // Uncomment this to support HTTPS.
+            // app.UseHttpsRedirection();
```
lzc850612, line 146: what is this line for? is .UseWebSockets() not enough?

lzc850612 (changes requested):

samples/csharp/skill/SkillSample/Adapters/DefaultAdapter.cs
gabog: AuthenticationConfiguration is initialized in statup.cs and adds the AllowedCallersClaimsValidator, makes sense?

```
// Register AuthConfiguration to enable custom claim validation.
services.AddSingleton(sp => new AuthenticationConfiguration { ClaimsValidator = new AllowedCallersClaimsValidator(sp.GetService<IConfiguration>()) });
```

samples/csharp/skill/SkillSample/Startup.cs
gabog: This is just builerplaite code from the Asp.Net core template, we normally leave it commented out in case the dev wants to enforce that only https requests are made to the web api (it is for HTTP not websockets).

More details [here](https://docs.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-3.1&tabs=visual-studio#require-https)

Do you want me to remove it?

lzc850612: right yeah. in that case we can keep it

lzc850612 (approved):

PR #3229, opened 2020-03-24
add oauth credentials to MultiProviderAuthDialog

lzc850612:
### Purpose

Add OAuthCredentials to MultiProviderAuthDialog to support using a different credentials for auth
use LG for MultiProviderAuthDialog
Remove ResponseManager and related classes

### Changes

### Tests

### Feature Plan

#### General


xieofxie:
Could we use lg in solution library now? since I think lg files will not be published to nuget?

sdk/csharp/libraries/microsoft.bot.solutions/Testing/BotTestBase.cs
```diff
@@ -1,60 +1,12 @@
...
 // Licensed under the MIT License.
 
-using System.Collections.Specialized;
-using System.Text.RegularExpressions;
-using Microsoft.Bot.Solutions.Responses;
+using System;
 
 namespace Microsoft.Bot.Solutions.Testing
 {
+    [Obsolete("This class is being deprecated. Please use Microsoft.Bot.Solutions.Responses.LocaleTemplateManager to handle response generation.")]
```
ryanisgrig, line 11: Is this the right message?

ryanisgrig: Nevermind, I see it was only used for ResponseManager

lzc850612: yeah a bit confusing but the functions are all for response generation

lzc850612:
> Could we use lg in solution library now? since I think lg files will not be published to nuget?

Developers should be able to use MultiProviderAuthDialog without having to passing a whole set of LG responses. That means we need to include the LG files. We should be able to do it by including them explicitly in the .nuspec file. I'm having trouble making the nuspec work now with azure build pipeline so still working on it (getting @darrenj to help)

xieofxie:
> > Could we use lg in solution library now? since I think lg files will not be published to nuget?
> 
> Developers should be able to use MultiProviderAuthDialog without having to passing a whole set of LG responses. That means we need to include the LG files. We should be able to do it by including them explicitly in the .nuspec file. I'm having trouble making the nuspec work now with azure build pipeline so still working on it (getting @darrenj to help)

Good to know the nuspec function! I once thought we could load content from assembly but it seems SDK doesn't expose ParseText in MultiLanguageLG

lauren-mills (approved):

lzc850612:
Sadly even if we did include the LG files in the nuget files, they still don't get copied over to the project. So i'm reverting back to use ResponseManager. Please take a look.

Issue #3215, opened 2020-03-18
Add a sample MVC Web Chat client using enhanced Direct Line features


ryanisgrig:
## User Story
As ... a developer
I want ... an out of the box web chat app enhanced w
so that ... I don't have to piece together my own web chat implementation

## Acceptance Criteria 
*List your requirements.*
- [ ] Base it off of the .NET MVC app used for the test environment (stored in internal Solutions repo)
- [ ] Enable Direct Line Speech per instructions from botframework-webchat repository: https://github.com/microsoft/BotFramework-WebChat/blob/master/docs/DIRECT_LINE_SPEECH.md
- [ ] Enable the enhanced Direct Line authentication
![image](https://user-images.githubusercontent.com/43043272/77009824-d666f700-6925-11ea-8a81-0f75a481e015.png)
    - Blog post: https://blog.botframework.com/2018/09/25/enhanced-direct-line-authentication-features/
    - Sample: https://github.com/microsoft/BotBuilder-Samples/tree/master/experimental/DirectLineTokenSite
- [ ] Provide documentation to go under the Clients and Channels section on how to run this sample

## Resources
*Provide any screenshots or conversation mockups to support your request.*
*Learn how to mock up bot conversations with [.chat files](http://aka.ms/bfchatfile).*

## Notes
*Provide any additional information.*

lzc850612:
Would documentation be enough? I image the Webchat code would have no difference from the official Webchat.

ryanisgrig:
Not really because if we are going to write the documentation for stringing various pieces together we should include a fully working sample out of the box. We also need this for our test environment anyways so may as well just share the code.

lzc850612:
is the client going to be essentially like this: https://blog.botframework.com/2018/09/01/using-webchat-with-azure-bot-services-authentication/

ryanisgrig:
It will use that and have Direct Line Speech integration

Sent from my 📱

________________________________

PR #3231, opened 2020-03-24
Direct Line Web Chat Sample

peterinnesmsft:
### Purpose
#3215: Adding a sample web chat client that integrates using either Direct Line or Direct Line Speech depending upon application configuration.

### Changes
Adding new sample client project, DirectLine.Web, to the Clients solution.

### Tests
No tests as this is sample code only. Manually verified using both configurations.

### Feature Plan
N/A

#### General


ryanisgrig:
@peterinnesmsft how did you validate the Enhanced DL flow locally? Did you have to publish a new web app to test against?

peterinnesmsft:
> @peterinnesmsft how did you validate the Enhanced DL flow locally? Did you have to publish a new web app to test against?

Thankfully I did not have to create a new web app. Within the Web App Bot resource, I configured the Direct Line Channel to enable Enhanced Auth - as part of this, for testing a local app, you need to add http://localhost as a Trusted Origin (N.B.: even though it might be locally hosted using HTTPS, you explicitly need to set using HTTP, otherwise the origin presented in the token to ABS does not successfully find the corresponding session which is required as part of the auth flow).

Beyond this, I also used ngrok against locally running versions of the Skill and VA to verify that the activity containing the access token from the OAuth Prompt was indeed successfully sent back to the Skill, by updating the Web App Bot configs to point to the temporary ngrok endpoints. Otherwise, that was the extent of it.

lauren-mills (approved):

samples/csharp/clients/directline.web/DirectLine.Web.csproj
```diff
@@ -0,0 +1,17 @@
...
+    <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="2.2.0" PrivateAssets="All" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="wwwroot\images\" />
```
lzc850612: probably not needed

lzc850612 (changes requested):

samples/csharp/clients/directline.web/DirectLine.Web.csproj
peterinnesmsft: Removed in latest iteration.

darrenj (approved):
We should have an accompanying doc item guiding how to make use of this sample and background.

lzc850612 (approved):

PR #3239, opened 2020-03-26
VirtualAssistantSample updates

gabog:
This PR contains several suggested changes to the VASkillSample based on the SDK samples and other observations:

- Refactored DefaultAdapter based on SDK Skill samples (broke down OnTurnError in several discrete methods with error handling and deleted conversation state at the end).
- Moved assignment of MainDialog.TelemetryClient into DefaultActivityHandler to avoid calling a virtual property from the Dialog's constructor.
- Updated MainDialog to store the active skill in convo state so we can send an EoC in OnTurnError if the VA crashes and is not handled.
- Reorganized startup.cs a bit and also refactored to use SkillsConfiguration for creating the skill dialog instances rather than rereading the appsettings and dublicating config keys that are already in SkillsConfiguration.
- Made private fields readonly where possible.
- Re-added some top level BF references to the SkillSample.csproj so devs can update SDK packages without having to update the Solutions package.
- Added missing cancellationToken parameters across the board.
- Added Async suffic to async methods.
- Updated program.cs based on OOTB 3.1 program.cs for asp.net core)

### DIFF
--- a/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/VirtualAssistantSample.FunctionalTests.csproj
+++ b/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/VirtualAssistantSample.FunctionalTests.csproj
@@ -16,7 +16,6 @@
     <PackageReference Include="Microsoft.Rest.ClientRuntime" Version="2.3.21" />
     <PackageReference Include="MSTest.TestAdapter" Version="2.0.0" />
     <PackageReference Include="MSTest.TestFramework" Version="2.0.0" />
-    <PackageReference Include="StyleCop.Analyzers" Version="1.1.118" PrivateAssets="all" />
   </ItemGroup>
 
   <ItemGroup>
--- a/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/VirtualAssistantSample.FunctionalTests.csproj
+++ b/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/VirtualAssistantSample.FunctionalTests.csproj
@@ -11,21 +11,16 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="AdaptiveCards" Version="1.2.4" />
-    <PackageReference Include="Microsoft.Bot.Builder.LanguageGeneration" Version="4.8.1-rc0" />
     <PackageReference Include="Microsoft.Bot.Connector.DirectLine" Version="3.0.2" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.4.0" />
     <PackageReference Include="Microsoft.Rest.ClientRuntime" Version="2.3.21" />
     <PackageReference Include="MSTest.TestAdapter" Version="2.0.0" />
     <PackageReference Include="MSTest.TestFramework" Version="2.0.0" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.1.118" PrivateAssets="all" />
   </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\VirtualAssistantSample.Tests\VirtualAssistantSample.Tests.csproj" />
     <ProjectReference Include="..\VirtualAssistantSample\VirtualAssistantSample.csproj" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <PackageReference Update="StyleCop.Analyzers" Version="1.1.118" />
   </ItemGroup>
 </Project>
--- a/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Adapters/DefaultAdapter.cs
+++ b/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Adapters/DefaultAdapter.cs
@@ -1,37 +1,61 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.Bot.Builder;
 using Microsoft.Bot.Builder.Azure;
 using Microsoft.Bot.Builder.Integration.ApplicationInsights.Core;
 using Microsoft.Bot.Builder.Integration.AspNet.Core;
+using Microsoft.Bot.Builder.Integration.AspNet.Core.Skills;
+using Microsoft.Bot.Builder.Skills;
+using Microsoft.Bot.Builder.TraceExtensions;
 using Microsoft.Bot.Connector.Authentication;
 using Microsoft.Bot.Schema;
 using Microsoft.Bot.Solutions.Feedback;
 using Microsoft.Bot.Solutions.Middleware;
 using Microsoft.Bot.Solutions.Responses;
+using Microsoft.Bot.Solutions.Skills;
+using Microsoft.Extensions.Logging;
+using VirtualAssistantSample.Dialogs;
 using VirtualAssistantSample.Services;
 
 namespace VirtualAssistantSample.Adapters
 {
     public class DefaultAdapter : BotFrameworkHttpAdapter
     {
+        private readonly ConversationState _conversationState;
+        private readonly ILogger _logger;
+        private readonly IBotTelemetryClient _telemetryClient;
+        private readonly LocaleTemplateManager _templateEngine;
+        private readonly SkillHttpClient _skillClient;
+        private readonly SkillsConfiguration _skillsConfig;
+        private readonly BotSettings _settings;
+
         public DefaultAdapter(
             BotSettings settings,
             ICredentialProvider credentialProvider,
             IChannelProvider channelProvider,
-            LocaleTemplateManager templateFile,
+            AuthenticationConfiguration authConfig,
+            LocaleTemplateManager templateEngine,
             ConversationState conversationState,
             TelemetryInitializerMiddleware telemetryMiddleware,
-            IBotTelemetryClient telemetryClient)
-            : base(credentialProvider, channelProvider)
+            IBotTelemetryClient telemetryClient,
+            ILogger<BotFrameworkHttpAdapter> logger,
+            SkillsConfiguration skillsConfig = null,
+            SkillHttpClient skillClient = null)
+            : base(credentialProvider, authConfig, channelProvider, logger: logger)
         {
-            OnTurnError = async (turnContext, exception) =>
-            {
-                await turnContext.SendActivityAsync(new Activity(type: ActivityTypes.Trace, text: $"Exception Message: {exception.Message}, Stack: {exception.StackTrace}"));
-                await turnContext.SendActivityAsync(templateFile.GenerateActivityForLocale("ErrorMessage", settings.DefaultLocale));
-                telemetryClient.TrackException(exception);
-            };
+            _conversationState = conversationState ?? throw new ArgumentNullException(nameof(conversationState));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _templateEngine = templateEngine ?? throw new ArgumentNullException(nameof(templateEngine));
+            _telemetryClient = telemetryClient ?? throw new ArgumentNullException(nameof(telemetryClient));
+            _skillClient = skillClient;
+            _skillsConfig = skillsConfig;
+            _settings = settings;
+
+            OnTurnError = HandleTurnErrorAsync;
 
             Use(telemetryMiddleware);
 
@@ -43,6 +67,80 @@ namespace VirtualAssistantSample.Adapters
             Use(new EventDebuggerMiddleware());
             Use(new FeedbackMiddleware(conversationState, telemetryClient, new FeedbackOptions()));
             Use(new SetSpeakMiddleware());
+        }
+
+        private async Task HandleTurnErrorAsync(ITurnContext turnContext, Exception exception)
+        {
+            // Log any leaked exception from the application.
+            _logger.LogError(exception, $"[OnTurnError] unhandled error : {exception.Message}");
+
+            await SendErrorMessageAsync(turnContext, exception);
+            await EndSkillConversationAsync(turnContext);
+            await ClearConversationStateAsync(turnContext);
+        }
+
+        private async Task SendErrorMessageAsync(ITurnContext turnContext, Exception exception)
+        {
+            try
+            {
+                _telemetryClient.TrackException(exception);
+
+                // Send a message to the user.
+                await turnContext.SendActivityAsync(_templateEngine.GenerateActivityForLocale("ErrorMessage"));
+
+                // Send a trace activity, which will be displayed in the Bot Framework Emulator.
+                // Note: we return the entire exception in the value property to help the developer;
+                // this should not be done in production.
+                await turnContext.TraceActivityAsync("OnTurnError Trace", exception.ToString(), "https://www.botframework.com/schemas/error", "TurnError");
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, $"Exception caught in SendErrorMessageAsync : {ex}");
+            }
+        }
+
+        private async Task EndSkillConversationAsync(ITurnContext turnContext)
+        {
+            if (_skillClient == null || _skillsConfig == null)
+            {
+                return;
+            }
+
+            try
+            {
+                // Inform the active skill that the conversation is ended so that it has a chance to clean up.
+                // Note: the root bot manages the ActiveSkillPropertyName, which has a value while the root bot
+                // has an active conversation with a skill.
+                var activeSkill = await _conversationState.CreateProperty<BotFrameworkSkill>(MainDialog.ActiveSkillPropertyName).GetAsync(turnContext, () => null);
+                if (activeSkill != null)
+                {
+                    var endOfConversation = Activity.CreateEndOfConversationActivity();
+                    endOfConversation.Code = "RootSkillError";
+                    endOfConversation.ApplyConversationReference(turnContext.Activity.GetConversationReference(), true);
+
+                    await _conversationState.SaveChangesAsync(turnContext, true);
+                    await _skillClient.PostActivityAsync(_settings.MicrosoftAppId, activeSkill, _skillsConfig.SkillHostEndpoint, (Activity)endOfConversation, CancellationToken.None);
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, $"Exception caught on attempting to send EndOfConversation : {ex}");
+            }
+        }
+
+        private async Task ClearConversationStateAsync(ITurnContext turnContext)
+        {
+            try
+            {
+                // Delete the conversationState for the current conversation to prevent the
+                // bot from getting stuck in a error-loop caused by being in a bad state.
+                // ConversationState should be thought of as similar to "cookie-state" for a Web page.
+                await _conversationState.DeleteAsync(turnContext);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, $"Exception caught on attempting to Delete ConversationState : {ex}");
+            }
         }
     }
 }
--- a/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Bots/DefaultActivityHandler.cs
+++ b/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Bots/DefaultActivityHandler.cs
@@ -22,13 +22,14 @@ namespace VirtualAssistantSample.Bots
         private readonly Dialog _dialog;
         private readonly BotState _conversationState;
         private readonly BotState _userState;
-        private IStatePropertyAccessor<DialogState> _dialogStateAccessor;
-        private IStatePropertyAccessor<UserProfileState> _userProfileState;
-        private LocaleTemplateManager _templateManager;
+        private readonly IStatePropertyAccessor<DialogState> _dialogStateAccessor;
+        private readonly IStatePropertyAccessor<UserProfileState> _userProfileState;
+        private readonly LocaleTemplateManager _templateManager;
 
         public DefaultActivityHandler(IServiceProvider serviceProvider, T dialog)
         {
             _dialog = dialog;
+            _dialog.TelemetryClient = serviceProvider.GetService<IBotTelemetryClient>();
             _conversationState = serviceProvider.GetService<ConversationState>();
             _userState = serviceProvider.GetService<UserState>();
             _dialogStateAccessor = _conversationState.CreateProperty<DialogState>(nameof(DialogState));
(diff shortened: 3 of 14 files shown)


samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Startup.cs
```diff
@@ -180,7 +165,11 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
                 .UseStaticFiles()
                 .UseWebSockets()
                 .UseRouting()
+                .UseAuthorization()
```
lzc850612: what would this add?

gabog: Leftover from the standard asp.net core template, we probably don't need it because we don't use ASP.Net core auth, removed it.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/VirtualAssistantSample.FunctionalTests.csproj
```diff
@@ -11,21 +11,16 @@
...
 
   <ItemGroup>
-    <PackageReference Include="AdaptiveCards" Version="1.2.4" />
-    <PackageReference Include="Microsoft.Bot.Builder.LanguageGeneration" Version="4.8.1-rc0" />
     <PackageReference Include="Microsoft.Bot.Connector.DirectLine" Version="3.0.2" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.4.0" />
     <PackageReference Include="Microsoft.Rest.ClientRuntime" Version="2.3.21" />
     <PackageReference Include="MSTest.TestAdapter" Version="2.0.0" />
     <PackageReference Include="MSTest.TestFramework" Version="2.0.0" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.1.118" PrivateAssets="all" />
```
lauren-mills: I don't think we should include the stylecop packages in the csproj for these samples since we export directly from these to update the templates. Can you add it to the build.props file instead?

lauren-mills (changes requested):

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/VirtualAssistantSample.FunctionalTests.csproj
gabog: I put it there because I saw you guys have them in the SkillSample too, should I remove it from all the projects?

gabog: For samples, I am personally in favor of removing them from all the projects, SyleCop is more a decision a dev needs to make and will probably not include them in the csprojs or vsix, I can put them in build.props so out code looks consistent but it is not forced on people that clone them.

lzc850612: i think it's fine to add it here. IMHO it's actually better since the developers could use this for stylecop purposes, and the props files won't make it into the template.

lauren-mills: I don't think we should force stylecop into the templates. That's a decision devs need to make for themselves and it doesn't seem like a good fit.

gabog: I am with @lauren-mills on this one, is it OK if I just move it to build.props and remove it from all the csprojs? Here is the list of the projects that I will remove it from: 

- SkillSample.csproj
- SkillSample.Tests.csproj
- VirtualAssistantSample.csproj
- VirtualAssistantSample.FunctionalTests.csproj
- VirtualAssistantSample.Tests.csproj

gabog: done

lauren-mills (approved):

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/VirtualAssistantSample.FunctionalTests.csproj
lzc850612: can you make the stypecop version change (to use 1.1.118) to the skill sample's props file? https://github.com/microsoft/botframework-solutions/blob/acdfdda8b1223f602ed279d2b6227197e9105218/samples/csharp/skill/Directory.Build.props#L7

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Startup.cs
```diff
@@ -181,6 +166,9 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
                 .UseWebSockets()
                 .UseRouting()
                 .UseEndpoints(endpoints => endpoints.MapControllers());
+
+            // Uncomment this to support HTTPS.
+            // app.UseHttpsRedirection();
```
peterinnesmsft, line 171: I would naively think we should always want to enable HTTPS for non-local development by default. Any reason we wouldn't just do a check to configure this if !env.IsDevelopment() instead?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Startup.cs
```diff
@@ -104,68 +109,48 @@ public void ConfigureServices(IServiceCollection services)
...
 
-            // Configure bot
-            services.AddTransient<IBot, DefaultActivityHandler<MainDialog>>();
-
             // Register the skills conversation ID factory, the client and the request handler.
             services.AddSingleton<SkillConversationIdFactoryBase, SkillConversationIdFactory>();
             services.AddHttpClient<SkillHttpClient>();
             services.AddSingleton<ChannelServiceHandler, SkillHandler>();
 
+            // Register dialogs
```
peterinnesmsft, line 122: >// Register dialogs [](start = 12, length = 19)

Just for my own education, is there any particular order in which services need to be registered with DI?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Bots/DefaultActivityHandler.cs
```diff
@@ -22,13 +22,14 @@ public class DefaultActivityHandler<T> : TeamsActivityHandler
...
         private readonly BotState _conversationState;
         private readonly BotState _userState;
-        private IStatePropertyAccessor<DialogState> _dialogStateAccessor;
-        private IStatePropertyAccessor<UserProfileState> _userProfileState;
-        private LocaleTemplateManager _templateManager;
+        private readonly IStatePropertyAccessor<DialogState> _dialogStateAccessor;
+        private readonly IStatePropertyAccessor<UserProfileState> _userProfileState;
+        private readonly LocaleTemplateManager _templateManager;
 
         public DefaultActivityHandler(IServiceProvider serviceProvider, T dialog)
```
peterinnesmsft, line 29: >IServiceProvider serviceProvider, T dialog [](start = 38, length = 42)

Do we need ArgumentNullException checks on these?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -27,25 +27,25 @@ namespace VirtualAssistantSample.Dialogs
...
+        private readonly SkillsConfiguration _skillsConfig;
+        private readonly LocaleTemplateManager _templateManager;
+        private readonly IStatePropertyAccessor<UserProfileState> _userProfileState;
+        private readonly IStatePropertyAccessor<List<Activity>> _previousResponseAccessor;
+        private readonly IStatePropertyAccessor<BotFrameworkSkill> _activeSkillProperty;
 
         public MainDialog(
-            IServiceProvider serviceProvider,
-            IBotTelemetryClient telemetryClient)
+            IServiceProvider serviceProvider)
```
peterinnesmsft, line 43: >IServiceProvider serviceProvider [](start = 12, length = 32)

Same thing here, arg null check?

peterinnesmsft, line 43: >IServiceProvider serviceProvider) [](start = 12, length = 33)

nit: this can probably move to the same line as above

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -77,6 +77,9 @@ public class MainDialog : ComponentDialog
             {
                 AddDialog(dialog);
             }
+
+            // Create state property to track the active skill.
+            _activeSkillProperty = conversationState.CreateProperty<BotFrameworkSkill>(ActiveSkillPropertyName);
```
peterinnesmsft: Consider moving this above to keep grouped with _previousResponseAccessor?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -145,15 +148,15 @@ protected override async Task<DialogTurnResult> OnContinueDialogAsync(DialogCont
...
+            innerDc.Context.OnSendActivities(StoreOutgoingActivitiesAsync);
             return await base.OnContinueDialogAsync(innerDc, cancellationToken);
         }
 
         private async Task<bool> InterruptDialogAsync(DialogContext innerDc, CancellationToken cancellationToken)
         {
             var interrupted = false;
             var activity = innerDc.Context.Activity;
-            var userProfile = await _userProfileState.GetAsync(innerDc.Context, () => new UserProfileState());
+            var userProfile = await _userProfileState.GetAsync(innerDc.Context, () => new UserProfileState(), cancellationToken);
```
peterinnesmsft, line 159: >var [](start = 12, length = 3)

Change from var to explicit type - I'm assuming this is returning UserProfileState, but it's a little hard for me to be sure since I'm less familiar with the code.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -168,11 +171,10 @@ private async Task<bool> InterruptDialogAsync(DialogContext innerDc, Cancellatio
                 // Check if we need to switch skills.
                 if (isSkill && IsSkillIntent(dispatchIntent) && dispatchIntent.ToString() != dialog.Id && dispatchScore > 0.9)
                 {
-                    EnhancedBotFrameworkSkill identifiedSkill;
-                    if (_skillsConfig.Skills.TryGetValue(dispatchIntent.ToString(), out identifiedSkill))
+                    if (_skillsConfig.Skills.TryGetValue(dispatchIntent.ToString(), out var identifiedSkill))
```
peterinnesmsft, line 174: >var [](start = 88, length = 3)

Explicit type

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -193,17 +195,17 @@ private async Task<bool> InterruptDialogAsync(DialogContext innerDc, Cancellatio
                         {
                             case GeneralLuis.Intent.Cancel:
                                 {
-                                    await innerDc.Context.SendActivityAsync(_templateManager.GenerateActivityForLocale("CancelledMessage", userProfile));
-                                    await innerDc.CancelAllDialogsAsync();
-                                    await innerDc.BeginDialogAsync(InitialDialogId);
+                                    await innerDc.Context.SendActivityAsync(_templateManager.GenerateActivityForLocale("CancelledMessage", userProfile), cancellationToken);
+                                    await innerDc.CancelAllDialogsAsync(cancellationToken);
```
peterinnesmsft, line 199: >CancelAllDialogsAsync [](start = 50, length = 21)

It seems like the idea is that we should always pass in the cancellation token. Do we need to consider removing the function prototypes that make it optional? Admittedly that's a breaking change to the SDK, but it would seem that making it optional could otherwise lead to undesirable behavior.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -193,17 +195,17 @@ private async Task<bool> InterruptDialogAsync(DialogContext innerDc, Cancellatio
...
+                                    await innerDc.BeginDialogAsync(InitialDialogId, cancellationToken: cancellationToken);
```
peterinnesmsft, line 200: >cancellationToken:  [](start = 84, length = 19)

Do we need parameter name explicitly here? Seems unnecessary.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -303,7 +305,7 @@ private async Task<DialogTurnResult> IntroStepAsync(WaterfallStepContext stepCon
         private async Task<DialogTurnResult> RouteStepAsync(WaterfallStepContext stepContext, CancellationToken cancellationToken)
         {
             var activity = stepContext.Context.Activity.AsMessageActivity();
-            var userProfile = await _userProfileState.GetAsync(stepContext.Context, () => new UserProfileState());
+            var userProfile = await _userProfileState.GetAsync(stepContext.Context, () => new UserProfileState(), cancellationToken);
```
peterinnesmsft, line 308: >var [](start = 12, length = 3)

Explicit type

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -319,70 +321,74 @@ private async Task<DialogTurnResult> RouteStepAsync(WaterfallStepContext stepCon
                     var dispatchIntentSkill = dispatchIntent.ToString();
                     var skillDialogArgs = new BeginSkillDialogOptions { Activity = (Activity)activity };
 
+                    // Save active skill in state.
+                    var selectedSkill = _skillsConfig.Skills[dispatchIntentSkill];
```
peterinnesmsft, line 325: >var [](start = 20, length = 3)

Explicit type

peterinnesmsft, line 325: >_skillsConfig.Skills[dispatchIntentSkill]; [](start = 40, length = 42)

Is this always guaranteed to exist? Worried about KeyNotFoundException potential.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -319,70 +321,74 @@ private async Task<DialogTurnResult> RouteStepAsync(WaterfallStepContext stepCon
...
+                    await _activeSkillProperty.SetAsync(stepContext.Context, selectedSkill, cancellationToken);
+
                     // Start the skill dialog.
-                    return await stepContext.BeginDialogAsync(dispatchIntentSkill, skillDialogArgs);
+                    return await stepContext.BeginDialogAsync(dispatchIntentSkill, skillDialogArgs, cancellationToken);
                 }
-                else if (dispatchIntent == DispatchLuis.Intent.q_Faq)
+
+                if (dispatchIntent == DispatchLuis.Intent.q_Faq)
```
peterinnesmsft, line 332: >dispatchIntent == DispatchLuis.Intent.q_Faq [](start = 20, length = 43)

What's the general practice on case sensitivity for matching intents? Is is typical to be sensitive?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -319,70 +321,74 @@ private async Task<DialogTurnResult> RouteStepAsync(WaterfallStepContext stepCon
...
+                await dc.CancelAllDialogsAsync(cancellationToken);
             }
             else
             {
                 throw new InvalidOperationException("OAuthPrompt.SignOutUser(): not supported by the current adapter");
             }
         }
 
-        private async Task<ResourceResponse[]> StoreOutgoingActivities(ITurnContext turnContext, List<Activity> activities, Func<Task<ResourceResponse[]>> next)
+        private async Task<ResourceResponse[]> StoreOutgoingActivitiesAsync(ITurnContext turnContext, List<Activity> activities, Func<Task<ResourceResponse[]>> next)
```
peterinnesmsft, line 391: >StoreOutgoingActivitiesAsync [](start = 47, length = 28)

Looks like we need cancellationToken wired through here as well if I'm not mistaken?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/OnboardingDialog.cs
```diff
@@ -18,13 +18,12 @@ namespace VirtualAssistantSample.Dialogs
...
-        private LocaleTemplateManager _templateManager;
-        private IStatePropertyAccessor<UserProfileState> _accessor;
+        private readonly BotServices _services;
+        private readonly LocaleTemplateManager _templateManager;
+        private readonly IStatePropertyAccessor<UserProfileState> _accessor;
 
         public OnboardingDialog(
-            IServiceProvider serviceProvider,
-            IBotTelemetryClient telemetryClient)
+            IServiceProvider serviceProvider)
```
peterinnesmsft, line 26: >IServiceProvider serviceProvider) [](start = 12, length = 33)

Move to same line as above?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Adapters/DefaultAdapter.cs
```diff
@@ -1,37 +1,61 @@
...
+using Microsoft.Bot.Solutions.Skills;
+using Microsoft.Extensions.Logging;
+using VirtualAssistantSample.Dialogs;
 using VirtualAssistantSample.Services;
 
 namespace VirtualAssistantSample.Adapters
 {
     public class DefaultAdapter : BotFrameworkHttpAdapter
     {
+        private readonly ConversationState _conversationState;
```
peterinnesmsft, line 28: >_conversationState [](start = 43, length = 18)

Yay for alpha-sorting! :) #Closed

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Startup.cs
gabog: That would be a dev decision I think, we try not to be too opinionated on our packages but I would defer the final decision on weather to do this to @lauren-mills and @lzc850612.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Adapters/DefaultAdapter.cs
```diff
@@ -44,5 +68,79 @@ public class DefaultAdapter : BotFrameworkHttpAdapter
...
+            try
+            {
+                // Inform the active skill that the conversation is ended so that it has a chance to clean up.
+                // Note: the root bot manages the ActiveSkillPropertyName, which has a value while the root bot
+                // has an active conversation with a skill.
+                var activeSkill = await _conversationState.CreateProperty<BotFrameworkSkill>(MainDialog.ActiveSkillPropertyName).GetAsync(turnContext, () => null);
+                if (activeSkill != null)
+                {
+                    var endOfConversation = Activity.CreateEndOfConversationActivity();
+                    endOfConversation.Code = "RootSkillError";
```
peterinnesmsft, line 118: >"RootSkillError" [](start = 45, length = 16)

Do we need to consider having some const definitions of codes like this somewhere?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Adapters/DefaultAdapter.cs
```diff
@@ -44,5 +68,79 @@ public class DefaultAdapter : BotFrameworkHttpAdapter
...
+                    endOfConversation.ApplyConversationReference(turnContext.Activity.GetConversationReference(), true);
```
peterinnesmsft, line 119: >true [](start = 114, length = 4)

named property

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Adapters/DefaultAdapter.cs
```diff
@@ -44,5 +68,79 @@ public class DefaultAdapter : BotFrameworkHttpAdapter
...
+
+                    await _conversationState.SaveChangesAsync(turnContext, true);
```
peterinnesmsft, line 121: >true [](start = 75, length = 4)

named property

peterinnesmsft (dismissed):
Looks good. Left comments mostly on styling things.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Startup.cs
lzc850612: I agree. leaving this commented so that developers can easily enable it later on when they want to, so the non-https wouldn't just not work when it's deployed to azure.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Startup.cs
lzc850612: I think the orders won't matter. i think they're all lazy-load

gabog: Not that this point, my understanding is that the DI services try gets built later on in the app life cycle (once you try to use IServiceProvider for example). There are some cases where you may need to access a services to configure another one, in those cases I created a local instance that I register and also use in the code (e.g.: BotSettings).
The DI guidance is not very explicit about the order, I'll try to find some better documentation, this is what I found so far: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-3.1#the-configure-method)

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Bots/DefaultActivityHandler.cs
lzc850612: don't think these two need them. serviceProvider is always there, and generic type parameter should be also otherwise there'll be an exception trying to resolve the dependencies of this class in DI phase

gabog: In general we do, across the board but didn't see that in the rest of the code so I just followed what was there already. @lauren-mills / @lzc850612, do we need this now? or we can do another PR for it?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
gabog: Maybe, pending @lauren-mills and @lzc850612 confirmation.

gabog: Just following the current coding style, I personally like constructor params in the same line unless the list is too long, but I will defer to @lauren-mills / @lzc850612 on this.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
gabog: Done

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
gabog: The general recommendation is to use var instead explicit unless needed. Some info here: https://docs.microsoft.com/en-us/visualstudio/ide/reference/convert-var-to-explicit-type?view=vs-2019

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
gabog: The general recommendation is to use var instead explicit unless needed. Some info here: https://docs.microsoft.com/en-us/visualstudio/ide/reference/convert-var-to-explicit-type?view=vs-2019

peterinnesmsft (approved):

PR #3277, opened 2020-04-08
[Botskills] Implement intent filtering

### LINKED ISSUE #3137: Provide a way to specify which intents to include out of an .LU file in the botskills CLI
## User Story
As a user, I want to be able to specify which intents to extract out of an `.lu` file when adding a skill as a dispatch source using `botskills connect`. This functionality is already supported by the `dispatch add` command which is internally being called by the `botskills connect`, but it is not exposed in the interface of `botskills connect`. 

My use case is as follows: I have a skill that I want to connect to, but only consume a limited number of its intents. The skill defines a large number of intents, including Confirm/Cancel/Reject,.... which I don't want to include in my dispatcher. One solution would be to define another `.lu` file with just the intents I want to connect to, but then synchronization between the two `.lu` files would be an issue. 

This picking of intents to extract can be done in one of 2 ways as far as I can see:
- Check which intents are exposed in the manifest
- Add the `--includedIntents` parameter to `botskills connect` and link it to `dispatch add`.

Related Ticket: https://github.com/microsoft/botframework-solutions/issues/2488

dfavretto:
Fixes #3137

### Purpose
Implement intent filtering when connecting a Skill with specific intents specified in the Skill's manifest.

### Changes
Retrieve the list of intent from the manifest and pass it to the dispatch using the `--includedIntents` argument.
If the manifest contains an intent with the name `*`, ignore the `--includedIntents` argument.

### Tests
New tests were added to cover this new functionality.

### Feature Plan

#### General


darrenj:
Looks good, could we update the docs in another PR just to help guide people. Thanks!

dfavretto:
> 
> 
> Looks good, could we update the docs in another PR just to help guide people. Thanks!

Sure! 😊

darrenj (approved):

darrenj:
@dfavretto - conflicts!

lzc850612 (approved):

Batta32:
@darrenj - we already fixed the conflicts! 😊

PR #3288, opened 2020-04-16
[TypeScript][Virtual Assistant & Skill] Add lowerCaseLng config to i18next

### LINKED ISSUE #3253: Personal Assistant Project will only work on windows
#### What project is affected
Virtual Assistant
node.js
#### What happens?
when i18next loads it looks for language files like this: en-US.json etc.  However, the project has them named like this: en-us.json.  That works ok in windows because window file access is case-insensitive.  However, it breaks in linux in a very frustrating way because none of the languages load etc.

Run node.js app in linux container
Default app works.

Batta32:
Fixes #3253 

### Purpose
Makes `i18next` use **lowercase language codes** to avoid case sensitiviness issues.

### Changes
Adds a configuration option for i18next to use language codes in lower case. This configuration is replicated to the Virtual Assistant and Skill Template and Sample

### Tests
\-

### Feature Plan
\-

#### General

### DIFF
--- a/templates/typescript/generator-botbuilder-assistant/generators/app/templates/sample-assistant/src/index.ts
+++ b/templates/typescript/generator-botbuilder-assistant/generators/app/templates/sample-assistant/src/index.ts
@@ -48,6 +48,7 @@ import { skills as skillsRaw } from './skills.json';
 // Configure internationalization and default locale
 i18next.use(i18nextNodeFsBackend)
     .init({
+        lowerCaseLng: true,
         fallbackLng: 'en-us',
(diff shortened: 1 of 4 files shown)


genbod:
This looks great.  It would be great to see some testing on both linux and windows node.js apps.  This bug would have been caught if the code was tested running in linux and windows.

Batta32:
Thanks @genbod, we will take into account your suggestion :blush:.

darrenj (approved):

darrenj (approved):

peterinnesmsft (approved):

PR #3291, opened 2020-04-16
Chitchat as fallback

### LINKED ISSUE #3233: QnAMaker Personalities cause Dispatcher Bias
Virtual Assistant Core

All

#### What happens?
After deploying Virtual Assistant, if you load one of the Chit Chat Personality Datasets (https://github.com/Microsoft/BotBuilder-PersonalityChat/tree/master/CSharp/Datasets), the dispatcher will have bias that causes top level intents to route to Chit Chat over more appropriate dialogs such as KB / Skills.

1. Create a VA deployment

1. Load the Chit Chat dataset from https://github.com/Microsoft/BotBuilder-PersonalityChat/tree/master/CSharp/Datasets

1. Run Dispatch update process

Expectation is that QnA Maker Chit Chat would not produce bias and cause intent routing issues for other skills or KB items

taicchoumsft:
Fixes [#3233](https://github.com/microsoft/botframework-solutions/issues/3233)
### Purpose
LUIS has a tendency to overfit on training data. Our personality datasets tend be very big, and once introduced at the dispatch level, tends to cause dispatch to be biased towards the chitchat intent. 

This PR moves chitchat out of dispatch, and calls the chitchat kb only after skills, faq and other intents are evaluated, and a user-adjustable threshold is met.

### Changes
The suggested threshold and conditionals are heuristics.  Might have to be adjusted after some testing.

### Tests
 \-

### Feature Plan
 \-

#### General

### DIFF
--- a/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
+++ b/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
@@ -338,7 +338,7 @@ namespace VirtualAssistantSample.Dialogs
                     return await stepContext.BeginDialogAsync(knowledgebaseId, cancellationToken: cancellationToken);
                 }
 
-                if (dispatchIntent == DispatchLuis.Intent.q_Chitchat)
+                if (ShouldBeginChitChatDialog(stepContext, dispatchIntent, dispatchScore))
                 {
                     stepContext.SuppressCompletionMessage(true);
 
@@ -440,13 +440,48 @@ namespace VirtualAssistantSample.Dialogs
         {
             if (dispatchIntent.ToString().Equals(DispatchLuis.Intent.l_General.ToString(), StringComparison.InvariantCultureIgnoreCase) ||
                 dispatchIntent.ToString().Equals(DispatchLuis.Intent.q_Faq.ToString(), StringComparison.InvariantCultureIgnoreCase) ||
-                dispatchIntent.ToString().Equals(DispatchLuis.Intent.q_Chitchat.ToString(), StringComparison.InvariantCultureIgnoreCase) ||
                 dispatchIntent.ToString().Equals(DispatchLuis.Intent.None.ToString(), StringComparison.InvariantCultureIgnoreCase))
             {
                 return false;
             }
 
             return true;
+        }
+
+        /// <summary>
+        /// A simple set of heuristics to govern if we should invoke the personality <see cref="QnAMakerDialog"/>.
+        /// </summary>
+        /// <param name="stepContext">Current dialog context</param>
+        /// <param name="dispatchIntent">Intent that Dispatch thinks should be invoked</param>
+        /// <param name="dispatchScore">Confidence score for intent</param>
+        /// <param name="threshold">User provided threshold, if above this threshold do NOT show chitchat</param>
+        /// <returns>A <see cref="bool"/> indicating if we should invoke the personality dialog</returns>
+        private bool ShouldBeginChitChatDialog(WaterfallStepContext stepContext, DispatchLuis.Intent dispatchIntent, double dispatchScore, double threshold = 0.5)
+        {
+            if (IsSkillIntent(dispatchIntent))
+            {
+                return false;
+            }
+            else if (dispatchIntent == DispatchLuis.Intent.None)
+            {
+                return true;
+            }
+            else if (dispatchIntent == DispatchLuis.Intent.l_General)
+            {
+                // If dispatch classifies user query as general, we should check against the cached general Luis score instead.
+                var generalResult = stepContext.Context.TurnState.Get<GeneralLuis>(StateProperties.GeneralResult);
+                if (generalResult != null)
+                {
+                    (var _, var generalScore) = generalResult.TopIntent();
+                    return generalScore < threshold;
+                }
+            }
+            else if (dispatchScore < threshold)
+            {
+                return true;
+            }
+
+            return false;
         }
     }
 }
--- a/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
+++ b/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
@@ -451,16 +451,16 @@ namespace VirtualAssistantSample.Dialogs
         /// <summary>
         /// A simple set of heuristics to govern if we should invoke the personality <see cref="QnAMakerDialog"/>.
         /// </summary>
-        /// <param name="stepContext">Current dialog context</param>
-        /// <param name="dispatchIntent">Intent that Dispatch thinks should be invoked</param>
-        /// <param name="dispatchScore">Confidence score for intent</param>
-        /// <param name="threshold">User provided threshold, if above this threshold do NOT show chitchat</param>
-        /// <returns>A <see cref="bool"/> indicating if we should invoke the personality dialog</returns>
+        /// <param name="stepContext">Current dialog context.</param>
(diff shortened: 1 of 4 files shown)


samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -440,13 +440,48 @@ private bool IsSkillIntent(DispatchLuis.Intent dispatchIntent)
...
+        /// <param name="dispatchScore">Confidence score for intent</param>
+        /// <param name="threshold">User provided threshold, if above this threshold do NOT show chitchat</param>
+        /// <returns>A <see cref="bool"/> indicating if we should invoke the personality dialog</returns>
+        private bool ShouldBeginChitChatDialog(WaterfallStepContext stepContext, DispatchLuis.Intent dispatchIntent, double dispatchScore, double threshold = 0.5)
+        {
+            if (IsSkillIntent(dispatchIntent))
+            {
+                return false;
+            }
+            else if (dispatchIntent == DispatchLuis.Intent.None)
```
peterinnesmsft: >else [](start = 12, length = 4)

You can ditch this else - entering into this block will return, so no need to have an else statement. #Resolved

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -440,13 +440,48 @@ private bool IsSkillIntent(DispatchLuis.Intent dispatchIntent)
...
+            {
+                return true;
+            }
+            else if (dispatchIntent == DispatchLuis.Intent.l_General)
```
peterinnesmsft: >else [](start = 12, length = 4)

This one can also be ditched. #Resolved

peterinnesmsft:
I would naively assume that these changes need to be made in the templates as well? Also do we need to do this for Skills?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -440,13 +440,48 @@ private bool IsSkillIntent(DispatchLuis.Intent dispatchIntent)
...
+            {
+                // If dispatch classifies user query as general, we should check against the cached general Luis score instead.
+                var generalResult = stepContext.Context.TurnState.Get<GeneralLuis>(StateProperties.GeneralResult);
```
ryanisgrig, line 474: The GeneralResult is how l_General scored in the Dispatch app correct?

ryanisgrig: Or GeneralResult is the score in the General LUIS app?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -440,13 +440,48 @@ private bool IsSkillIntent(DispatchLuis.Intent dispatchIntent)
...
+        /// A simple set of heuristics to govern if we should invoke the personality <see cref="QnAMakerDialog"/>.
+        /// </summary>
+        /// <param name="stepContext">Current dialog context</param>
+        /// <param name="dispatchIntent">Intent that Dispatch thinks should be invoked</param>
+        /// <param name="dispatchScore">Confidence score for intent</param>
+        /// <param name="threshold">User provided threshold, if above this threshold do NOT show chitchat</param>
+        /// <returns>A <see cref="bool"/> indicating if we should invoke the personality dialog</returns>
+        private bool ShouldBeginChitChatDialog(WaterfallStepContext stepContext, DispatchLuis.Intent dispatchIntent, double dispatchScore, double threshold = 0.5)
+        {
+            if (IsSkillIntent(dispatchIntent))
```
ryanisgrig: `IsSkillIIntent` is called before the `ShouldBeginChitChatDialog`, why does it need to be checked here? #Resolved

taicchoumsft: Yes, skill intent and many others are checked before we reach this point - this is added just in case we change the flow of dialogs. Will remove if you feel this is unncessary.

---
In reply to: [410416659](https://github.com/microsoft/botframework-solutions/pull/3291#discussion_r410416659) [](ancestors = 410416659)

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
taicchoumsft: GeneralResult is the score in the General LUIS app.  What was found during testing was that dispatch might score a dispatch level intent as l_general with very high confidence.  But if we pass the same intent to the General LUIS app, the score is significantly lower.  So in this case it's better to take the General LUIS app's score as the final arbiter.

taicchoumsft:
Discussed this with Patrick yesterday, and he mentioned templates is generated from sample when we roll out new vsix templates.  No changes needed here. For skills, they do not have chitchat at all, so this is a VA level only change. 

---
In reply to: [615398125](https://github.com/microsoft/botframework-solutions/pull/3291#issuecomment-615398125) [](ancestors = 615398125)

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
lzc850612: yeah i think we should remove it as judging by the name of this function it shouldn't care about SkillIntent. #Resolved

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Dialogs/MainDialog.cs
```diff
@@ -440,13 +440,50 @@ private bool IsSkillIntent(DispatchLuis.Intent dispatchIntent)
...
+
+        /// <summary>
+        /// A simple set of heuristics to govern if we should invoke the personality <see cref="QnAMakerDialog"/>.
+        /// </summary>
+        /// <param name="stepContext">Current dialog context</param>
+        /// <param name="dispatchIntent">Intent that Dispatch thinks should be invoked</param>
+        /// <param name="dispatchScore">Confidence score for intent</param>
+        /// <param name="threshold">User provided threshold, if above this threshold do NOT show chitchat</param>
+        /// <returns>A <see cref="bool"/> indicating if we should invoke the personality dialog</returns>
+        private bool ShouldBeginChitChatDialog(WaterfallStepContext stepContext, DispatchLuis.Intent dispatchIntent, double dispatchScore, double threshold = 0.5)
```
lzc850612, line 459: to make the code more robust, if you want to make the threshold as a parameter, be sure to check if threshold is within the range of 0 to 1.0 #Resolved

lzc850612 (changes requested):

lzc850612:
If you are making the change to by default disable ChitChat intent, should we remove q_ChitChat from DispatchLuis.cs? @lauren-mills what do you think? #Resolved

lzc850612 (approved):

pavolum (approved):

darrenj:
Great thanks for testing this and getting the change in for the next release. @Batta32 - Could we take this change into typescript VA please ahead of our release?

darrenj:
@peterinnesmsft @taicchoumsft For our release docs we should have a set of steps for customers to follow with existing VAs to adopt this change. e.g:

- dispatch remove
- dispatch refresh
- change code
- Update deployment scripts (link to existing docs for this)

PR #3299, opened 2020-04-21
Pavolum/telephony changes

pavolum:
### Purpose
Adding the following components needed to support telephony channel out of the box
- Added code to ignore blank message activities from DirectLineSpeech channel within activityHandler for both speech and VA
- Added code to grab speak values from attachments if the activity does not have a text or speak value and the attachment has a speak value
- Changes voice for speech scenarios to more human like neural voice
- Added speak values to HeroCard attachment types

### DIFF
--- a/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Bots/DefaultActivityHandler.cs
+++ b/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Bots/DefaultActivityHandler.cs
@@ -7,7 +7,9 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Bot.Builder;
 using Microsoft.Bot.Builder.Dialogs;
+using Microsoft.Bot.Builder.Dialogs.Choices;
 using Microsoft.Bot.Builder.Teams;
+using Microsoft.Bot.Connector;
 using Microsoft.Bot.Schema;
 using Microsoft.Bot.Solutions;
 using Microsoft.Bot.Solutions.Responses;
@@ -66,6 +68,13 @@ namespace VirtualAssistantSample.Bots
 
         protected override Task OnMessageActivityAsync(ITurnContext<IMessageActivity> turnContext, CancellationToken cancellationToken)
         {
+            // directline speech occasionally sends empty message activities that should be ignored
+            var activity = turnContext.Activity;
+            if (activity.ChannelId == Channels.DirectlineSpeech && activity.Type == ActivityTypes.Message && string.IsNullOrEmpty(activity.Text))
+            {
+                return Task.CompletedTask;
+            }
+
             return _dialog.RunAsync(turnContext, _dialogStateAccessor, cancellationToken);
         }
 
--- a/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Responses/MainResponses.lg
+++ b/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Responses/MainResponses.lg
@@ -58,8 +58,12 @@
 
 # EscalateMessage
 [HeroCard
-  subtitle=Our agents are available 24/7 at 1(800)555-1234. Or connect with us through Microsoft Teams.
+  subtitle= ${EscalatedText()}
+  speak = ${EscalatedText()}
 ]
+
+# EscalatedText
+- Our agents are available 24/7 at 1(800)555-1234. Or connect with us through Microsoft Teams.
 
 # NewUserIntroCard
 [Activity
@@ -206,8 +210,12 @@
 # HelpCard
 [HeroCard 
     title=Help for Virtual Assistant
-    subtitle=This card can be used to display information to help your user interact with your bot.
+    subtitle=${HelpText()}
+    speak = ${HelpText()}
 ]
+
+# HelpText
+- This can be used to display information to help your user interact with your bot.
 
 # RandomName 
 - IF: ${Name && rand(0, 1000) > 500}
--- a/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Responses/MainResponses.lg
+++ b/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Responses/MainResponses.lg
@@ -215,7 +215,7 @@
 ]
 
 # HelpText
-- This can be used to display information to help your user interact with your bot.
+- This can be used to display information to help your user interact with your Virtual Assistant.
 
 # RandomName 
 - IF: ${Name && rand(0, 1000) > 500}
--- a/sdk/csharp/libraries/microsoft.bot.solutions/Middleware/SetSpeakMiddleware.cs
+++ b/sdk/csharp/libraries/microsoft.bot.solutions/Middleware/SetSpeakMiddleware.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Bot.Solutions.Middleware
         /// Initializes a new instance of the <see cref="SetSpeakMiddleware"/> class.
         /// </summary>
         /// <param name="locale">If null, use en-US.</param>
-        /// <param name="voiceFonts">Map voice font for locale like en-US to "Microsoft Server Speech Text to Speech Voice (en-US, ZiraRUS)".</param>
+        /// <param name="voiceFonts">Map voice font for locale like en-US to "Microsoft Server Speech Text to Speech Voice (en-US, Jessa24kRUS)".</param>
         /// <param name="channels">Set SSML for these channels. If null, use <see cref="Connector.Channels.DirectlineSpeech"/> and <see cref="Connector.Channels.Emulator"/>.</param>
         public SetSpeakMiddleware(string locale = null, IDictionary<string, string> voiceFonts = null, ISet<string> channels = null)
         {
@@ -111,7 +111,7 @@ namespace Microsoft.Bot.Solutions.Middleware
                 var attachmentContent = activity.Attachments[0].Content;
                 if (attachmentContent != null && attachmentContent.GetType() == typeof(JObject))
                 {
-                    var contentObj = (JObject)attachmentContent;
+                    var contentObj = attachmentContent as JObject;
                     return contentObj["speak"]?.ToString();
                 }
             }
--- a/sdk/csharp/libraries/microsoft.bot.solutions/Middleware/SetSpeakMiddleware.cs
+++ b/sdk/csharp/libraries/microsoft.bot.solutions/Middleware/SetSpeakMiddleware.cs
@@ -12,6 +12,7 @@ namespace Microsoft.Bot.Solutions.Middleware
     using System.Xml.Linq;
     using Microsoft.Bot.Builder;
     using Microsoft.Bot.Schema;
+    using Newtonsoft.Json.Linq;
 
     /// <summary>
     /// Set Speech Synthesis Markup Language (SSML) on an Activity's Speak property with locale and voice input.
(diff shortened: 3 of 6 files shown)


darrenj (changes requested):
Looks like Jessa has been deprecated in favour of Aria - as per [this](https://docs.microsoft.com/en-us/azure/cognitive-services/speech-service/language-support#neural-voices). Could we take this opportunity to replace all instances of jessa with Aria and quickly validate?

pavolum:
> Looks like Jessa has been deprecated in favour of Aria - as per [this](https://docs.microsoft.com/en-us/azure/cognitive-services/speech-service/language-support#neural-voices). Could we take this opportunity to replace all instances of jessa with Aria and quickly validate?

I saw that deprecation message before and tried loads of different ways but no matter what, when I switch it to Aria the bot does not speak and I get no meaningful error messages. I am going to follow up with Egor and Autumn and see if they have any suggestions.

pavolum:
> Looks like Jessa has been deprecated in favour of Aria - as per [this](https://docs.microsoft.com/en-us/azure/cognitive-services/speech-service/language-support#neural-voices). Could we take this opportunity to replace all instances of jessa with Aria and quickly validate?

This is why I should reach out to others before spending loads of time trying to debug something that gives no error messages... I was just advised that Neural voices are not supported in all data centers. Here is a doc showing where it is supported: https://docs.microsoft.com/en-us/azure/cognitive-services/speech-service/regions#standard-and-neural-voices

This explains why my attempts to switch to the neural voices led to the bot not talking at all. 

Since we can't guarantee which region folks deploy VAs to, we will need to stick with standard voices in the template, I am going to keep the change to the Jessa24kRUS voice since I feel it sounds more human like then the previous one we had.

ryanisgrig:
> > Looks like Jessa has been deprecated in favour of Aria - as per [this](https://docs.microsoft.com/en-us/azure/cognitive-services/speech-service/language-support#neural-voices). Could we take this opportunity to replace all instances of jessa with Aria and quickly validate?
> 
> This is why I should reach out to others before spending loads of time trying to debug something that gives no error messages... I was just advised that Neural voices are not supported in all data centers. Here is a doc showing where it is supported: https://docs.microsoft.com/en-us/azure/cognitive-services/speech-service/regions#standard-and-neural-voices
> 
> This explains why my attempts to switch to the neural voices led to the bot not talking at all.
> 
> Since we can't guarantee which region folks deploy VAs to, we will need to stick with standard voices in the template, I am going to keep the change to the Jessa24kRUS voice since I feel it sounds more human like then the previous one we had.

I think that voice change makes sense for the template. FYI this will reflect on the test environments so if we want a neural voice for demos this is another differentiator we have to account for.

sdk/csharp/libraries/microsoft.bot.solutions/Middleware/SetSpeakMiddleware.cs
```diff
@@ -91,11 +92,39 @@ public Task OnTurnAsync(ITurnContext context, NextDelegate next, CancellationTok
...
+                return result;
+            }
+
+            // return speak value of first attachment if an attachment exists and has a speak value
+            if (activity.Attachments.Count > 0)
+            {
+                var attachmentContent = activity.Attachments[0].Content;
+                if (attachmentContent != null && attachmentContent.GetType() == typeof(JObject))
+                {
+                    var contentObj = (JObject)attachmentContent;
```
taicchoumsft: >(JObject)attachmentContent [](start = 37, length = 26)

minor nit: `attachmentContent as JObject` may be more C#-ish.  Then you can remove the `typeof` check above.

pavolum: Im not sure if Attachment.content will always be a JObject. The attachment class dictates it is of type Object so maybe there are scenarios in which is not a JObject. Changing the way I cast is fine, but removing the typeof check might be dangerous, right? Unless im missing something.

taicchoumsft: The `as` operator tries to convert the object to the type you requested, if the conversion is not possible, it returns `null`.  Since you do a null check right below, you should be fine removing the `typeof` check. 

---
In reply to: [413301899](https://github.com/microsoft/botframework-solutions/pull/3299#discussion_r413301899) [](ancestors = 413301899)

taicchoumsft (approved):
:shipit:

sdk/csharp/libraries/microsoft.bot.solutions/Middleware/SetSpeakMiddleware.cs
```diff
@@ -91,11 +92,39 @@ public Task OnTurnAsync(ITurnContext context, NextDelegate next, CancellationTok
...
+            }
+
+            // return speak value of first attachment if an attachment exists and has a speak value
+            if (activity.Attachments.Count > 0)
+            {
+                var attachmentContent = activity.Attachments[0].Content;
+                if (attachmentContent != null && attachmentContent.GetType() == typeof(JObject))
+                {
+                    var contentObj = attachmentContent as JObject;
+                    return contentObj["speak"]?.ToString();
```
taicchoumsft, line 115: >contentObj [](start = 27, length = 10)

question: Not sure if `speak` is also optional... in that case it might be safer to check that too - `contentObj?["speak"]?.ToString();`

taicchoumsft: >contentObj [](start = 27, length = 10)

Sorry, we missed the the null check for `contentObj`  - `contentObj?["speak"]?.ToString()`.

pavolum: Ahh I see, with this null check then the removal of getType makes sense

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Bots/DefaultActivityHandler.cs
```diff
@@ -66,6 +68,13 @@ protected override async Task OnMembersAddedAsync(IList<ChannelAccount> membersA
 
         protected override Task OnMessageActivityAsync(ITurnContext<IMessageActivity> turnContext, CancellationToken cancellationToken)
         {
+            // directline speech occasionally sends empty message activities that should be ignored
```
peterinnesmsft, line 71: >// directline speech occasionally sends empty message activities that should be ignored [](start = 12, length = 87)

This seems weird. Do we know DL Speech does this?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Responses/MainResponses.lg
```diff
@@ -206,9 +210,13 @@
 # HelpCard
 [HeroCard 
     title=Help for Virtual Assistant
-    subtitle=This card can be used to display information to help your user interact with your bot.
+    subtitle=${HelpText()}
+    speak = ${HelpText()}
 ]
 
+# HelpText
+- This can be used to display information to help your user interact with your bot.
```
peterinnesmsft: >bot [](start = 79, length = 3)

To be consistent with the tile and general naming, change bot to Virtual Assistant?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Bots/DefaultActivityHandler.cs
lzc850612: i think i've run into this as well. It'd be good to follow up with DL Speech folks to see why this is the case. It could be that there's some bug in DL Speech

sdk/csharp/libraries/microsoft.bot.solutions/Middleware/SetSpeakMiddleware.cs
```diff
@@ -91,11 +92,39 @@ public Task OnTurnAsync(ITurnContext context, NextDelegate next, CancellationTok
...
+            var result = activity.Speak ?? activity.Text;
+            if (result != null)
+            {
+                return result;
+            }
+
+            // return speak value of first attachment if an attachment exists and has a speak value
+            if (activity.Attachments.Count > 0)
+            {
+                var attachmentContent = activity.Attachments[0].Content;
```
lzc850612, line 111: what's the scenario that has attachment content for text and speech response?

lzc850612 (changes requested):

sdk/csharp/libraries/microsoft.bot.solutions/Middleware/SetSpeakMiddleware.cs
peterinnesmsft: I think you can actually further simplify a little based on improvements to the `is` keyword since C# 7.0. I would need to double-check the behavior, but I *think* you could write as:

`if (activity.Attachments.Count > 0)
{
    if (activity.Attachments[0].Content is JObject contentObj)
    {
        return contentObj["speak"]?.ToString();
    }
}`

---
In reply to: [413314558](https://github.com/microsoft/botframework-solutions/pull/3299#discussion_r413314558) [](ancestors = 413314558,413301899)

peterinnesmsft (approved):
:shipit:

darrenj (approved):

sdk/csharp/libraries/microsoft.bot.solutions/Middleware/SetSpeakMiddleware.cs
pavolum: In our core VA this scenario exists with the help card, onboarding card and escalation card. These scenarios are activities that have no speak or text value but do have attachments with a speak or text value. At current state, nothing would render in a pure speech channel. With this change they would work. 

There were a few options to handle this. I believe this is the best. Instead of changing the root activity in the above scenarios to have speak values, grabbing the speak value from the first attachment ensures all existing VAs work without having to add speak value for all the VAs we support (enterprise, hospitality, skills, etc.). I also think the logic makes sense for general VA development. It ensure speak values of attachments are not ignored in speech scenarios and are instead used as a back up if a speak/test value does not exist on the root activity.

sdk/csharp/libraries/microsoft.bot.solutions/Middleware/SetSpeakMiddleware.cs
```diff
@@ -91,11 +92,39 @@ public Task OnTurnAsync(ITurnContext context, NextDelegate next, CancellationTok
...
+                if (attachmentContent != null)
+                {
+                    var contentObj = attachmentContent as JObject;
+                    return contentObj?["speak"]?.ToString();
```
lzc850612, line 115: i think this would barf if contentObj doesn't have a 'speak' key, even if you have the ? operator after it. probably gonna be 'given key does not exist' or something like it. you need to explicitly check if contentObj has 'speak' key.

lzc850612 (changes requested):

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Bots/DefaultActivityHandler.cs
pavolum: Ill figure out who to alert and will spin up an email thread with them

sdk/csharp/libraries/microsoft.bot.solutions/Middleware/SetSpeakMiddleware.cs
pavolum: I tested that edge case and if it does not have the speak key then ```contentObj?["speak"]?.ToString()``` returns null.

I thought it would be an issue as well when Tai suggested this but tested it and sure enough he was right :)

lzc850612 (approved):

PR #3422, opened 2020-05-27
Exchanging data with skills

### LINKED ISSUE #2456: Allow passing data from the skill back to the VA
## User Story
As a developer, I would like to be able to pass data back from the skill to the VA.

I would like to be able to pass data back from the skill back to the invoking VA when the skill is complete, in order (for example) to continue the conversation with another bot, with the information generated by the first bot.

My use case is the following:

I have the VA connected to 2 bots, the first bot is used to search for people and the 2nd bot is used to book meetings with those people. One of the functionalities of the search people bot is to enable the user to create a meeting with a searched person. That would require the search people bot to return a reference to the person with whom the meeting will be created, back to the VA. That is currently [not possible as far as I can tell](https://stackoverflow.com/questions/58120862/botframework-passing-data-from-skill-to-virtual-assistant), because all that the skill returns to the VA on completion is a boolean value. 

It would be great if something like the manifest/SkillContext/slots mechanism is available from skill -> VA instead of just for VA -> skill

darrenj:
Close #2456 

Added documentation to cover how a VA can pass data to a Skill and inversely retrieve data from a Skill supporting more complex "aggregated" experiences like "Your day today"


ryanisgrig (approved):
Made minor edits

PR #3497, opened 2020-06-30
Add TeamsReactionMiddleware for processing Teams Reaction

sidmore:
### Purpose
a. Add TeamsReactionMiddleware
b. Process Activity.Reaction if it is added to the reaction list

### Changes

### Tests
a. Implemented UT for TeamsReactionMiddleware

### Feature Plan

#### General


sdk/csharp/libraries/microsoft.bot.solutions/Feedback/TeamsReactionMiddleware.cs
```diff
@@ -0,0 +1,47 @@
...
+            if (turnContext.Activity.ChannelId.Equals(Channels.Msteams, StringComparison.OrdinalIgnoreCase)
+                && turnContext.Activity.Type != null
+                && turnContext.Activity.Type == ActivityTypes.MessageReaction)
+            {
+                IList<MessageReaction> reactionsAdded = turnContext.Activity.ReactionsAdded;
+                var record = new FeedbackRecord
+                {
+                    TeamsReaction = reactionsAdded[reactionsAdded.Count - 1].Type,
+                    Tag = TeamsReactionMiddleware.TeamsFeedback,
+                    Request = turnContext.Activity,
```
pavolum, line 37: What is the value of turnContext.Activity for an activity of ActivityType == MessageReaction. Does the Activity have some data pertaining to the text inside the bot response that was reacted to? 

I just want to ensure that this FeedbackRecord being logged has the necessary data so on the analytics side a conclusion can be made. 'i.e. bot said 'x' and user did/did not like it'

sdk/csharp/libraries/microsoft.bot.solutions/Feedback/FeedbackHelper.cs
```diff
@@ -0,0 +1,28 @@
...
+        public static void LogFeedback(FeedbackRecord record, IBotTelemetryClient botTelemetryClient)
+        {
+            var properties = new Dictionary<string, string>()
+            {
+                { nameof(FeedbackRecord.Tag), record.Tag },
+                { nameof(FeedbackRecord.Feedback), record.Feedback },
+                { nameof(FeedbackRecord.Comment), record.Comment },
+                { nameof(FeedbackRecord.Request.Text), record.Request?.Text },
+                { nameof(FeedbackRecord.Request.Id), record.Request?.Conversation.Id },
+                { nameof(FeedbackRecord.Request.ChannelId), record.Request?.ChannelId },
```
pavolum, line 22: There is no logging of record.TeamsReaction so when this is called from line 40 in the teamsReactionMiddleware the record.TeamsReaction value is ignored and not logged. 

Furthermore if you do make the change to log it as a new value then you should also make the changes in our PowerBI dashboard to grab that particular reaction value and render it as part of our existing feedback visualizations. Otherwise that telemetry will only exist in the raw logging data and users will have to manually configure the visualisation.

Another option is to log 'TeamsReaction' as 'FeedbackRecord.Feedback' or 'FeedbackRecord.Comment' which our powerBI dashboard already has transform logic for. This would be less work and should still provide the data analyst what is needed to make a conclusion on the particular success of a given bot message but its sort of hacky and you would have to append some string to the value to indicate that it is teamsReactionFeedback and not general Feedback.

I believe this bit of middleware will live longer then our 'temp' feedback solution. If you agree then making a dedicated visual/field in the existing visual in the resulting PowerBI dashboard may be worth the effort.

pavolum (commented):
Great PR, nice and concise :)

Couple general comments...

If this will be part of the next merge from next to master then we should update the public documentation in next to inform folks how to use the middelware so the documentation is merged to master with this change. I would add that to the feedback.md page we already have going in VA/Handbook and explain how it is tied in with the existing 'temp' feedback mechanism (theyre essentially totally decoupled which is ideal considering this middleware will have a longer life span then the temp feedback solution in the docs). However if you are updating the temp feedback mechanism this can wait as you will need to update that whole doc pending any changes you make to the mechanism as a whole.

Also the PowerBI dashboard has not been updated. So with this current change, users will be able to handle logging reactions but they will manually have to set up the aggregation and visual analytics of that feedback. In our PowerBI dashboard template there is a feedback tab were you could add a visualization for teams or piggyback off existing visualizations. I have outlined this a bit more in one of my comments. If you decide to update the powerBI dashboard, Tai and/or Ryan should be able to help guide you on the process as there is no documentation in our internal wiki (I would offer to help but they know more then me here). If you go that route, creating documentation outlining how to update the template would be useful for the next time somebody is making an update.

PR #3502, opened 2020-07-01
[C#] Function test enhancements for VirtualAssistantSample/SkillSample

ryanisgrig:
## Purpose
Close [AB#45781](https://fuselabs.visualstudio.com/566d503b-55c0-46b5-9e83-8a4efa7f9efa/_workitems/edit/45781)
Close [AB#45788](https://fuselabs.visualstudio.com/566d503b-55c0-46b5-9e83-8a4efa7f9efa/_workitems/edit/45788)

## Changes
* Introducing the TestBotClient approach used by SDK's functional tests for Direct Line driven testing for both VirtualAssistant and Skill sample bots
* Introduce EventActivityPrompt in SkillSample which allows for SkillDialog to be tried by a SampleAction event

### Virtual Assistant Sample
**GeneralTests**
* Assert Cancel intent is triggered
* Assert Escalate intent is triggered
* Assert Help intent is triggered
* Assert Logout intent is triggered
* Assert Repeat intent is triggered
* Assert StartOver intent is triggered

**GreetingTests**
* Assert new user greeting is triggered
* Assert returning user greeting is triggered
* Assert onboarding dialog extracts a person's name from complex string (my name is X)

**QnAMakerTests**
* Assert Chit-Chit KB is triggered on fallback
* Assert FAQ KB is triggered

**SkillSampleTests**
* Assert that a connected SkillSample bot is triggered 

### Skill Sample
**SampleActionTests**
* SampleDialog is triggered by event

**SampleDialogTests**
* Sample Dialog is triggered by user utterance

### General


samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/CoreTests.cs
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        /// <summary>
+        /// Assert that a new user is greeted and begins onboarding prompt.
+        /// </summary>
+        /// <param name="cancellationTokenSource">CancellationTokenSource.</param>
+        /// <param name="testBot">TestBotClient.</param>
+        /// <param name="useComplexInputWithName">Boolean flag to determine user response to onboarding prompt.</param>
+        /// <returns>Task.</returns>
+        public async Task Assert_New_User_Greeting(CancellationTokenSource cancellationTokenSource, TestBotClient testBot, bool useComplexInputWithName = false)
```
sidmore: This is a good way to introduce new asserts. Infact may be these bot specific asserts could be in a lib

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Configuration/EnvironmentBotTestConfiguration.cs
```diff
@@ -0,0 +1,39 @@
...
+        private const string DirectlineSecretKey = "DIRECTLINE";
+        private const string BotIdKey = "BOTID";
+
+        public EnvironmentBotTestConfiguration(string directLineSecretKey, string botIdKey, string userId = null)
+        {
+            // Load config from environment variables
+            DirectLineSecret = Environment.GetEnvironmentVariable(directLineSecretKey);
+            if (string.IsNullOrWhiteSpace(DirectLineSecret))
+            {
+                Assert.Inconclusive($"Environment variable '{directLineSecretKey}' not found.");
```
peterinnesmsft, line 20: I would typically avoid usage of Assert in a constructor since we're not necessarily testing something here. Maybe throw something like an ArgumentException with some kind of message indicating the specified environment variable was not found?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Configuration/EnvironmentBotTestConfiguration.cs
```diff
@@ -0,0 +1,39 @@
...
+            }
+
+            BotId = Environment.GetEnvironmentVariable(botIdKey);
+            if (string.IsNullOrWhiteSpace(BotId))
+            {
+                Assert.Inconclusive($"Environment variable '{botIdKey}' not found.");
+            }
+        }
+
+        public EnvironmentBotTestConfiguration()
```
peterinnesmsft, line 30: minor nit: typically tend to put parameterless constructors first.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Configuration/BotTestConfiguration.cs
```diff
@@ -0,0 +1,34 @@
...
+            if (string.IsNullOrEmpty(directLineSecret))
+            {
+                throw new ArgumentNullException(nameof(directLineSecret));
+            }
+
+            BotId = botId;
+            DirectLineSecret = directLineSecret;
+        }
+
+        public BotTestConfiguration()
```
peterinnesmsft, line 26: nit: empty constructor should go first

samples/csharp/skill/SkillSample.FunctionalTests/Configuration/BotTestConfiguration.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace SkillSample.FunctionalTests.Configuration
+{
+    public class BotTestConfiguration : IBotTestConfiguration
```
peterinnesmsft, line 8: Same comments as in the VirtualAssistantSample version.

samples/csharp/skill/SkillSample.FunctionalTests/Configuration/EnvironmentBotTestConfiguration.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace SkillSample.FunctionalTests.Configuration
+{
+    public class EnvironmentBotTestConfiguration : IBotTestConfiguration
```
peterinnesmsft, line 9: Same comments as in the VirtualAssistantSample version.

samples/csharp/skill/SkillSample.FunctionalTests/Configuration/IBotTestConfiguration.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace SkillSample.FunctionalTests.Configuration
+{
+    public interface IBotTestConfiguration
```
peterinnesmsft, line 6: Could these files go in a shared dll since they're identical?

samples/csharp/skill/SkillSample.Tests/SkillTestBase.cs
```diff
@@ -31,6 +31,17 @@ public class SkillTestBase : BotTestBase
 
         public LocaleTemplateManager TemplateEngine { get; set; }
 
+        protected Templates AllResponsesTemplates
+        {
+            get
+            {
+                var path = CultureInfo.CurrentUICulture.Name.ToLower() == "en-us" ?
```
peterinnesmsft: I would recommend using String.Equals with StringComparison.OrdinalIgnoreCase generally for string comparisons rather than a hard ==

samples/csharp/skill/SkillSample.sln
```diff
@@ -7,20 +7,40 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SkillSample", "SkillSample\
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SkillSample.Tests", "SkillSample.Tests\SkillSample.Tests.csproj", "{83C5C21E-1F3A-483E-8C4C-1FC61F1AB4E7}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SkillSample.FunctionalTests", "SkillSample.FunctionalTests\SkillSample.FunctionalTests.csproj", "{D47E4DBE-9510-418E-9582-F972F28C44C6}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug - NuGet Packages|Any CPU = Debug - NuGet Packages|Any CPU
```
peterinnesmsft: What is this?

samples/csharp/skill/SkillSample.sln
```diff
@@ -7,20 +7,40 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SkillSample", "SkillSample\
...
 		Debug|Any CPU = Debug|Any CPU
+		Documentation|Any CPU = Documentation|Any CPU
```
peterinnesmsft: And this? I'm probably just being ignorant but not seen these before. :)

samples/csharp/skill/SkillSample/Dialogs/MainDialog.cs
```diff
@@ -50,6 +50,18 @@ public class MainDialog : ComponentDialog
             AddDialog(_sampleAction);
         }
 
+        // Validate that activity is either a message with text or an event
+        protected async Task<bool> Validator(PromptValidatorContext<Activity> promptContext, CancellationToken cancellationToken)
```
peterinnesmsft: This feels a little weird to put this here - this feels like throwing test code in. What are we trying to do here?

samples/csharp/skill/SkillSample/Models/EventActivityPrompt.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace SkillSample.Models
+{
+    /// <summary>
+    /// EventActivityPrompt enables user to add a prompt to the stack that will optionally except Event Activities.
+    /// </summary>
+    public class EventActivityPrompt : ActivityPrompt
+    {
+        public EventActivityPrompt(string dialogId, PromptValidator<Activity> validator)
+        : base(dialogId, validator)
```
peterinnesmsft: indent

samples/csharp/skill/SkillSample/Models/EventActivityPrompt.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+        }
+
+        public async Task OnPromptNullContext(object options, CancellationToken cancellationToken = default(CancellationToken))
```
peterinnesmsft: Add "Async" suffix

samples/csharp/skill/SkillSample/Models/EventActivityPrompt.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            var opt = (PromptOptions)options;
+
+            // should throw ArgumentNullException
+            await OnPromptAsync(turnContext: null, state: null, options: opt, isRetry: false);
+        }
+
+        public async Task OnPromptNullOptions(DialogContext dc, CancellationToken cancellationToken = default(CancellationToken))
```
peterinnesmsft: Add "Async" suffix

samples/csharp/skill/SkillSample/Models/EventActivityPrompt.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        public EventActivityPrompt(string dialogId, PromptValidator<Activity> validator)
+        : base(dialogId, validator)
+        {
+        }
+
+        public async Task OnPromptNullContext(object options, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var opt = (PromptOptions)options;
+
```
peterinnesmsft, line 21: Again, this feels a little like adding test code, which feels a bit weird. We should try to avoid putting test-specific code in at this layer.

samples/csharp/skill/SkillSample.FunctionalTests/SampleDialogTests.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
```
peterinnesmsft: Whoops! :)

samples/csharp/skill/SkillSample.FunctionalTests/SampleDialogTests.cs
```diff
@@ -0,0 +1,46 @@
...
+    [TestCategory("SampleDialog")]
+    public class SampleDialogTests : SkillTestBase
+    {
+        [TestMethod]
+        public async Task Test_Utterance_SampleDialog()
+        {
+            await Assert_Utterance_Triggers_SkillDialog();
+        }
+
+        public async Task Assert_Utterance_Triggers_SkillDialog()
```
peterinnesmsft: Is this being called elsewhere? If not, any reason to separate this out into a separate function? Feels like this could just be in the test method and that would be OK.

samples/csharp/skill/SkillSample.FunctionalTests/SampleDialogTests.cs
```diff
@@ -0,0 +1,46 @@
...
+            var profileState = new { Name = SampleDialogUtterances.NamePromptResponse };
+            var introTextVariations = AllResponsesTemplates.ExpandTemplate("IntroText");
+            var namePromptTextVariations = AllResponsesTemplates.ExpandTemplate("NamePromptText");
+            var haveNameMessageTextVariations = AllResponsesTemplates.ExpandTemplate("HaveNameMessageText", profileState);
+
+            var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromMinutes(2));
+
+            var testBot = new TestBotClient(new EnvironmentBotTestConfiguration());
+
+            await testBot.StartConversation(cancellationTokenSource.Token);
```
peterinnesmsft: This should have Async suffix.

peterinnesmsft: And same for any functions here that are using await but missing it.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,242 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Bot.Connector.DirectLine;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using SkillSample.FunctionalTests.Configuration;
+
+namespace SkillSample.FunctionalTests.Bot
+{
+    public class TestBotClient
```
peterinnesmsft, line 22: Could we put this in a shared test library and overlap with the other one defined?

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                    else
+                    {
+                        throw new Exception("Failed to acquire directline token");
+                    }
+                }
+            }
+        }
+
+        public string GetUser()
```
peterinnesmsft: This feels like it should be implemented as a readonly property instead of a function.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,242 @@
...
+                        User = new { Id = this.user },
+                        TrustedOrigins = new string[]
+                        {
+                            OriginHeaderValue
+                        }
+                    }), Encoding.UTF8, "application/json");
+
+                using (var response = client.SendAsync(request).GetAwaiter().GetResult())
+                {
+                    if (response.IsSuccessStatusCode)
```
peterinnesmsft: I would avoid doing this in-line within the constructor. I would consider refactoring to pass in the direct line token through the constructor. If there is a need to refresh this, then some kind of background refresh mechanism may be a little safer.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,242 @@
...
+                        this.conversationId = JsonConvert.DeserializeObject<DirectLineToken>(body).ConversationId;
+
+                        // Create directline client from token
+                        this.directLineClient = new DirectLineClient(token);
+
+                        // From now on, we'll add an Origin header in directline calls, with
+                        // the trusted origin we sent when acquiring the token as value.
+                        directLineClient.HttpClient.DefaultRequestHeaders.Add(OriginHeaderKey, OriginHeaderValue);
+                    }
+                    else
```
peterinnesmsft: Invert the check on response.IsSuccessStatusCode - you can throw at that point and remove branch.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,242 @@
...
+                    {
+                        throw new Exception("Failed to acquire directline token");
```
peterinnesmsft: It would be better to return an exception that indicates the response returned so that error information isn't lost.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,242 @@
...
+
+            // Create a message activity with the input text.
+            var messageActivity = new Activity
+            {
+                From = new ChannelAccount(this.user),
+                Text = message,
+                Type = ActivityTypes.Message,
+            };
+
+            Console.WriteLine($"Sent to bot: {message}");
```
peterinnesmsft: We should avoid using Console.WriteLine for logging and instead lean into ILogger. One of the benefits of using an HttpClient provided by HttpClientFactory is that it includes message handlers that log a fair amount of request and response information. If we need to add custom logging, that would be the way to go.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,242 @@
...
+            }
+
+            var resourceResponses = new List<ResourceResponse>();
+
+            foreach (var message in messages)
+            {
+                resourceResponses.Add(await SendMessageAsync(message, cancellationToken));
+            }
+
+            return resourceResponses.ToArray();
```
peterinnesmsft: Any reason to convert explicitly to an array? Why not just return `Task<IList<ResourceResponse>>`?

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        public async Task StartConversation(CancellationToken cancellationToken = default(CancellationToken))
```
peterinnesmsft: Add Async suffixes throughout where missing

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,242 @@
...
+            Assert.IsTrue(messagesList.Any(m => m.Type == ActivityTypes.Message && m.Text.Contains(expected, StringComparison.OrdinalIgnoreCase)), $"Expected: {expected}");
+        }
+
+        public async Task AssertReplyOneOf(IEnumerable<object> expected, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var messages = await PollBotMessagesAsync(cancellationToken);
+            Assert.IsTrue(messages.Any(m => m.Type == ActivityTypes.Message && expected.Any(e => m.Text.Contains(e.ToString(), StringComparison.OrdinalIgnoreCase))));
+        }
+
+        public async Task<IEnumerable<Activity>> PollBotMessagesAsync(CancellationToken cancellationToken = default(CancellationToken))
```
peterinnesmsft, line 179: Unless we're dealing with proactive, I would assume that there should typically be an expected number of messages to be returned? Would it make sense to put an expected activity count so that we don't have to spin unnecessarily if we're received all expected activities?

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,242 @@
...
+            // Retrieve activities from directline
+            var activitySet = await directLineClient.Conversations.GetActivitiesAsync(conversationId, watermark, cancellationToken);
+            watermark = activitySet?.Watermark;
+
+            // Extract and return the activities sent from the bot.
+            return activitySet == null ? null : activitySet?.Activities?.Where(activity => activity.From.Id == this.config.BotId && activity.Type == ActivityTypes.Message);
+        }
+    }
+
+    public class DirectLineToken
```
peterinnesmsft, line 211: Do we not have definitions for this anywhere in the SDK?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,439 @@
...
+using System.Threading.Tasks;
+using AdaptiveCards;
+using Microsoft.Bot.Connector.DirectLine;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using VirtualAssistantSample.FunctionalTests.Configuration;
+
+namespace VirtualAssistantSample.FunctionalTests
+{
+    public class TestBotClient
```
peterinnesmsft, line 25: Definitely some overlap - let's move to a common test dll, I think that makes sense.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,439 @@
...
+            return activitySet == null ? null : activitySet?.Activities?.Where(activity => activity.From.Id == this.config.BotId && activity.Type == ActivityTypes.Message);
+        }
+
+        public void VerifyAdaptiveCard(IEnumerable<object> expected, Activity adaptiveCard, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (adaptiveCard == null)
+            {
+                throw new Exception("AdaptiveCard is null");
+            }
+            else if (adaptiveCard.Attachments == null)
```
peterinnesmsft: You can remove the else

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,439 @@
...
+            {
+                throw new Exception("AdaptiveCard.Attachments = null");
+            }
+
+            var card = JsonConvert.DeserializeObject<AdaptiveCard>(JsonConvert.SerializeObject(adaptiveCard.Attachments.FirstOrDefault().Content));
```
peterinnesmsft: What is the expected type of Content? If this is how it's received in the response, I would naively assume that its underlying type would be JToken or similar. If that's the case, you would be able to save a serialization/deserialization and cast as JToken, then to .ToObject<AdaptiveCard> or similar.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,439 @@
...
+        {
+            // Retrieve activities from directline
+            var activitySet = await directLineClient.Conversations.GetActivitiesAsync(conversationId, watermark, cancellationToken);
+            watermark = activitySet?.Watermark;
+
+            // Extract and return the activities sent from the bot.
+            return activitySet == null ? null : activitySet?.Activities?.Where(activity => activity.From.Id == this.config.BotId && activity.Type == ActivityTypes.Message);
+        }
+
+        public void VerifyAdaptiveCard(IEnumerable<object> expected, Activity adaptiveCard, CancellationToken cancellationToken = default(CancellationToken))
```
peterinnesmsft: AssertAdaptiveCard?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,439 @@
...
+            if (card.Speak == null)
+            {
+                throw new Exception("No speak received in Adaptive Card");
+            }
+
+            // Verify that the speak property matches one of expected replies
+            Assert.IsTrue(expected.Any(e => card.Speak.Contains(e.ToString(), StringComparison.OrdinalIgnoreCase)));
+        }
+
+        public void VerifyHeroCard(IEnumerable<object> expected, Activity heroCard, CancellationToken cancellationToken = default(CancellationToken))
```
peterinnesmsft: AssertHeroCard?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/GeneralTests.cs
```diff
@@ -0,0 +1,169 @@
...
+using VirtualAssistantSample.Models;
+using VirtualAssistantSample.Tests;
+using VirtualAssistantSample.Tests.Utterances;
+
+namespace VirtualAssistantSample.FunctionalTests
+{
+    [TestClass]
+    [TestCategory("FunctionalTests")]
+    [TestCategory("General")]
+    public class GeneralTests : BotTestBase
```
peterinnesmsft: GeneralTests gives me the wrong impression :) GeneralIntentTests or similar?

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/GeneralTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        /// <summary>
+        /// Assert that Escalate intent is working.
+        /// </summary>
+        private async Task Assert_General_Escalate()
+        {
+            var escalateMessageVariations = AllResponsesTemplates.ExpandTemplate("EscalatedText");
+
+            var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromMinutes(2));
```
peterinnesmsft: Would it make sense to put the cancellation token source and test bot initialization into a TestInitialize function so you don't have to repeat it within each test case?

peterinnesmsft (changes requested):
Looks good. Would be good to make a few structural changes and move some classes to a common location.

samples/csharp/skill/SkillSample.sln
ryanisgrig: Seems there are added by Visual Studio. I'll try removing and see if the solution works the same.

samples/csharp/skill/SkillSample/Dialogs/MainDialog.cs
ryanisgrig: The text prompt here only listens for Message Activities, but the Skill Sample itself needs to demonstrate that a Skill can be triggered by either a sample utterance (message) or sample action (event). This validator function is a basic example of doing so.

samples/csharp/skill/SkillSample/Models/EventActivityPrompt.cs
ryanisgrig: Not test code, but validating the Sample Utterance/Action logic explained in an earlier comment.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
ryanisgrig: This method isn't even used, deleting.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
ryanisgrig: Method isn't used here, removing.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
ryanisgrig: Adding expected activity account logic could end up having scaling problems, my experience with running these functional tests in a batch is occasionally there will be a hiccup when we are looking at specific activities and their ordering. Instead, this method lets the develop just capture all of the returns bot messages and do with them as they wish.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
ryanisgrig: No, and in fact this `TestBotClient` class is implemented from the SDK samples.

ryanisgrig:
Thanks @peterinnesmsft, I've implemented most changes and left comments on the others. As far as moving some of the classes to a common location are you suggesting moving them into the Solutions library? That is something I considered but then we would need to publish another minor release to keep that accounted for. Additionally, there is some logic that we would want to leave up to the developer that we shouldn't just include in a library (i.e. asserting adaptive/hero cards have the correct content).

As far as a shared library between the two projects, these projects are packaged in the VA and Skill VSIX templates, meaning they would need their own copy of the classes to do so.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
peterinnesmsft: Spoke offline - given that these need to be delivered independently in the VSIX, we agreed that for now we should keep distinct copies given the challenges of having to publish NuGets and consume them, etc.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,434 @@
...
+                .AddLogging(builder =>
+                {
+                    builder.AddDebug();
+                    builder.AddConsole();
+                })
+                .BuildServiceProvider();
+
+            var factory = serviceProvider.GetService<ILoggerFactory>();
+
+            logger = factory.CreateLogger<TestBotClient>();
```
peterinnesmsft: nit: this. (and thank you for using that prefix because it really helps with readability for me!)

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,434 @@
...
+
+            // Instead of generating a vanilla DirectLineClient with secret,
+            // we obtain a directline token with the secrets and then we use
+            // that token to create the directline client.
+            // What this gives us is the ability to pass TrustedOrigins when obtaining the token,
+            // which tests the enhanced authentication.
+            // This endpoint is unfortunately not supported by the directline client which is
+            // why we add this custom code.
+            using (HttpClient client = new HttpClient())
```
peterinnesmsft: We can use var here

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,434 @@
...
+            {
+                HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, $"https://directline.botframework.com/v3/directline/tokens/generate");
```
peterinnesmsft: var

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,434 @@
...
+                request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", config.DirectLineSecret);
+                request.Content = new StringContent(
+                    JsonConvert.SerializeObject(new
+                {
+                    User = new { Id = this.user },
+                    TrustedOrigins = new string[]
+                        {
+                            OriginHeaderValue
+                        }
```
peterinnesmsft, line 89: Fix indentation

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,434 @@
...
+            //      3.a- Detect the PostSignIn url in the redirect chain
+            //      3.b- Add cookie and challenge session id to post sign in link
+            // 4- Verify final redirect to token service ends up in success
+
+            // 1- Verify we have a sign in link in the activity
+            if (oauthCard == null)
+            {
+                throw new Exception("OAuthCard is null");
+            }
+            else if (oauthCard.Attachments == null)
```
peterinnesmsft, line 284: No need for else since you're throwing in the above condition block.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,434 @@
...
+            }
+
+            var card = JsonConvert.DeserializeObject<SigninCard>(JsonConvert.SerializeObject(oauthCard.Attachments.FirstOrDefault().Content));
+
+            if (card == null)
+            {
+                throw new Exception("No SignIn Card received in activity");
+            }
+
+            if (card.Buttons == null || !card.Buttons.Any())
```
peterinnesmsft, line 296: Any should typically only be used to determine if a collection contains one or more items that match certain conditional criteria. If you just want to see if it's empty, it's more performant to just do .Length/.Count == 0

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,434 @@
...
+            {
+                throw new Exception("No buttons received in sign in card");
+            }
+
+            string signInUrl = card.Buttons[0].Value?.ToString();
+
+            if (string.IsNullOrEmpty(signInUrl) || !signInUrl.StartsWith("https://"))
```
peterinnesmsft: In general, unless case sensitivity matters, it's probably worthwhile to do case insensitive comparison.

samples/csharp/assistants/virtual-assistant/VirtualAssistantSample.FunctionalTests/Bot/TestBotClient.cs
```diff
@@ -0,0 +1,434 @@
...
+                AllowAutoRedirect = false,
+                CookieContainer = cookieContainer
+            };
+
+            // We have a sign in url, which will produce multiple HTTP 302 for redirects
+            // This will path
+            //      token service -> other services -> auth provider -> token service (post sign in)-> response with token
+            // When we receive the post sign in redirect, we add the cookie passed in the directline session info
+            // to test enhanced authentication. This in ther scenarios happens by itself since browsers do this for us.
+            using (var client = new HttpClient(handler))
```
peterinnesmsft, line 330: We should avoid creating a new HttpClient for each individual request, as it's generally more costly. It's better to just instantiate and reuse a client for the lifetime of the class, and then just dispose during cleanup of the class by implementing IDisposable.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
peterinnesmsft: Per offline discussion, avoiding shared library for now due to complexities in how these need to be surfaced in the VSIX templates.

samples/csharp/skill/SkillSample.FunctionalTests/Bot/TestBotClient.cs
peterinnesmsft: Woop woop! 🙃

samples/csharp/skill/SkillSample/Dialogs/MainDialog.cs
peterinnesmsft: Got it. Maybe rename to ValidateActivityAsync in that case?

peterinnesmsft (approved):
Will mark myself as non-blocking. Overall looks good. Would be good to address just some minor style things, but OK I'm fine with the changes here.

lzc850612 (approved):

lauren-mills:
Closing - outdated

PR #3534, opened 2020-07-07
[PORT #3152][Botskills] Get the LU filename using basename

Batta32:
Port #3152 

### Purpose
The use of `split('\\')` is incompatible between OS, [basename](https://nodejs.org/api/path.html#path_path_basename_path_ext) method returns the last portion of a path, similar to the Unix basename command ignoring the trailing directory separators.

### Changes
- Change the `split('\\')` to `basename` in order to get the last portion of path

### Tests
\-

### Feature Plan

#### General


taicchoumsft (approved):
:shipit:

ryanisgrig (approved):

Issue #3290, opened 2020-04-16
Implement cross-language functional test pipelines for Virtual Assistant and Skill samples


ryanisgrig:
## User Story
As ... a developer
I want ... YAML pipelines available for the Virtual Assistant and Skill Sample
so that ... I can run nightly validation tests against my VA solution

## Acceptance Criteria 
- [ ] Build off of investments in the [Skills Functional Tests](https://github.com/microsoft/BotFramework-FunctionalTests/blob/master/SkillsFunctionalTests/README.md) with YAML files available for the below cross-language scenarios. This should use the same task variables as the source files.

|Virtual Assistant/Skill|C#|TypeScript|
-|-|-
C#|✅|✅|
TypeScript|✅|✅|

### Prepare
* Access Azure Key Vault (AKV) secrets for VA & Skill
* If they exist, cleanup existing test resources groups
* Install dispatch, luis-api
* Install botframework-cli
* Install botskills
    * Optional variable to configure between stable or preview version for above npm packages
* Get BotBuilder version
    * Optional variable to configure between stable or preview version
* Get BotSolutions version
    * Optional variable to configure between stable or preview version

#### Virtual Assistant
* Update to custom package references
* Run dotnet restore
* Run dotnet build

#### Skill
* Update to custom package references
* Run dotnet restore
* Run dotnet build

#### Virtual Assistant
* Run unit tests

#### Skill
* Run unit tests

#### Virtual Assistant
* Run deployment script
    * Languages: en-us,de-de,fr-fr,it-it,es-es,zh-cn

#### Skill
* Run deployment script
    * Languages: en-us, de-de,fr-fr,it-it,es-es,zh-cn

#### Virtual Assistant
* Create Direct Line Channel registration
* Get channel secrets
* Run functional Direct Line tests

### Cleanup
* Delete resource group

- [ ] Update docs with additional overview on how to enable this (can be lifted from existing Skills Functional Tests doc)

## Resources
*Provide any screenshots or conversation mockups to support your request.*
*Learn how to mock up bot conversations with [.chat files](http://aka.ms/bfchatfile).*

## Notes
*Provide any additional information.*

peterinnesmsft:
Closing issue to track in internal backlog going forward.

PR #3463, opened 2020-06-12
[TypeScript][Virtual Assistant] Add Functional Tests

Batta32:
Related to #3290 

### Purpose
*What is the context of this pull request? Why is it being done?*
There is no functional test applied for Virtual Assistant in TypeScript.

### Changes
*Are there any changes that need to be called out as significant or particularly difficult to grasp? (Include illustrative screenshots for context if applicable.)*
- Copy and adapt C# VA functional test as the responses' length between C# and TS are different
- Add yml file for the TypeScript VA

### Tests
*Is this covered by existing tests or new ones? If no, why not?*
\-

### Feature Plan
*Are there any remaining steps or dependencies before this issue can be fully resolved? If so, describe and link to any relevant pull requests or issues.*
\-

#### General
- [ ] I have commented my code, particularly in hard-to-understand areas	
- [X] I have added or updated the appropriate tests	
- [ ] I have updated related documentation

PR #3542, opened 2020-07-08
[TypeScript][Skill] Add Functional Tests

Batta32:
Related to #3290 

### Purpose
There is no functional test applied for Skill in TypeScript.

### Changes
- Copy and adapt C# Skill functional test as the responses' length between C# and TS are different
- Add yml file for the TypeScript Skill

### Tests
\-

### Feature Plan
\-

#### General
- [X] I have updated related documentation


lzc850612 (approved):

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+using Microsoft.Bot.Connector.DirectLine;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace SkillSample.FunctionalTests
+{
+    [TestClass]
+    [TestCategory("FunctionalTests")]
+    public class DirectLineClientTests
+    {
+        private static readonly string TestName = "Jane Doe";
```
peterinnesmsft: Use const instead of static readonly

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+        private static readonly string TriggerSkill = "Run Sample dialog";
+
+        private static string _directLineSecret = string.Empty;
```
peterinnesmsft: Do these need to be static? If there's no particular need for these to be static, I would just make them regular instance member variables.

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            get
+            {
+                var path = CultureInfo.CurrentUICulture.Name.ToLower() == "en-us" ?
+                    Path.Combine(".", "Responses", $"AllResponses.lg") :
+                    Path.Combine(".", "Responses", $"AllResponses.{CultureInfo.CurrentUICulture.Name.ToLower()}.lg");
+                return Templates.ParseFile(path);
+            }
+        }
+
+        [TestInitialize]
```
peterinnesmsft: This logic looks like it only needs to be performed once, but using [TestInitialize] means it gets executed at the start of each test. Did you mean to use [ClassInitialize] instead?

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+
+            return botResponses;
+        }
+
+        /// <summary>
+        /// Get the values for the environment variables.
+        /// </summary>
+        private void GetEnvironmentVars()
+        {
+            if (string.IsNullOrWhiteSpace(_directLineSecret) || string.IsNullOrWhiteSpace(_botId))
```
peterinnesmsft: Isn't this always going to inherently be true that we need to pull in environment variables? Not sure I understand the value of this check.

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            {
+                _directLineSecret = Environment.GetEnvironmentVariable("DIRECTLINE");
+                if (string.IsNullOrWhiteSpace(_directLineSecret))
+                {
+                    Assert.Inconclusive("Environment variable 'DIRECTLINE' not found.");
```
peterinnesmsft: Avoid using an Assert for this. Throw an ArgumentNullException instead. I would also just fold this logic into a ClassInitialize function rather than have a separate function.

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+        private static string _directLineSecret = string.Empty;
+        private static string _botId = string.Empty;
+        private static DirectLineClient _client;
+        private static string _userID;
+
+        protected static Templates AllResponsesTemplates
+        {
+            get
+            {
+                var path = CultureInfo.CurrentUICulture.Name.ToLower() == "en-us" ?
```
peterinnesmsft: Use String.Equals(..., ..., StringComparison.OrdinalIgnoreCase) instead.

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+                    Path.Combine(".", "Responses", $"AllResponses.lg") :
+                    Path.Combine(".", "Responses", $"AllResponses.{CultureInfo.CurrentUICulture.Name.ToLower()}.lg");
```
peterinnesmsft: Store in a local variable rather than perform twice?

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+        {
+            var introText = AllResponsesTemplates.ExpandTemplate("IntroText");
+            var allFirstPromptTextVariations = AllResponsesTemplates.ExpandTemplate("FirstPromptText");
+
+            var conversation = await StartBotConversationAsync();
+
+            var responses = await SendActivityAsync(conversation, CreateStartConversationEvent(fromUser));
+
+            Assert.AreEqual(introText[0], responses[0].Text);
+            if (responses.Count == 3)
```
peterinnesmsft: I'm struggling to understand what the purpose of this check is for. Is this to handle for specific locales?

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+                Text = activityText,
+                Type = ActivityTypes.Message
+            };
+        }
+
+        /// <summary>
+        /// Starts a conversation with a bot.
+        /// </summary>
+        /// <returns>Returns the new conversation.</returns>
+        private static async Task<Conversation> StartBotConversationAsync()
```
peterinnesmsft: Do we really need a function for this? It feels a little unnecessary given that it's not simply wrapping the call to the client - it doesn't really save anything.

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            return await _client.Conversations.StartConversationAsync();
+        }
+
+        /// <summary>
+        /// Sends an activity and waits for the response.
+        /// </summary>
+        /// <returns>Returns the bots answer.</returns>
+        private static async Task<List<Activity>> SendActivityAsync(Conversation conversation, Activity activity)
+        {
+            var allFirstPromptTextVariations = AllResponsesTemplates.ExpandTemplate("FirstPromptText");
```
peterinnesmsft: This is being generated in the test method. Any reason not to pass what's computed there into here? Otherwise you run the risk of code going out of sync.

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+                responses = await ReadBotMessagesAsync(_client, conversation.ConversationId);
+
+                attempts++;
+
+                if (attempts == 5)
+                {
+                    throw new NullReferenceException("There was an error reading the response. It might be an issue with DirectLine.");
+                }
+
+                // In some TypeScript cases the first element is an allFirstPromptTextVariations' text which is an error, so we repeat the request.
```
peterinnesmsft: Why are we getting an error? If there is periodically an error occurring, then we should fix it so that this behavior is consistent, no?

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+                if (responses?.FirstOrDefault()?.Text == null || allFirstPromptTextVariations.Contains(responses?.FirstOrDefault()?.Text))
+                {
+                    // Wait for ten second before polling the bot again.
```
peterinnesmsft: Injecting arbitrary delays like this is not generally a good idea. What are we trying to accomplish here?

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+
+            // Poll the bot for replies once per second.
+
+            // Retrieve the activity sent from the bot.
+            var activitySet = await client.Conversations.GetActivitiesAsync(conversationId, watermark);
+            watermark = activitySet?.Watermark;
+
+            // Extract the activities sent from the bot.
+            var activities = from x in activitySet.Activities
+                             where x.From.Id == _botId
```
peterinnesmsft: Linq statements are a little slow in performance. Use Linq extension methods instead, e.g.:

List<Activity> botResponses = activitySet.Activities
    .Where(a => string.Equals(_botId, a.From?.Id, StringComparison.OrdinalIgnoreCase))
    .ToList();

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+                             select x;
+
+            // Analyze each activity in the activity set.
+            if (activities.Any())
```
peterinnesmsft: This is a bad usage of Any. Any should be used to determine presence of items that match one or more criteria in a collection. If you are simply checking to see whether the collection is empty, check Count/Length == 0.

That said, performing ToList on the IEnumerable is a relatively trivial thing to do here, so performing the above code in my previous suggestion should be fine.

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+
+        /// <summary>
+        /// Polls the bot continuously until it gets a response.
+        /// </summary>
+        /// <param name="client">The Direct Line client.</param>
+        /// <param name="conversationId">The conversation ID.</param>
+        /// <returns>Returns the bot's answer.</returns>
+        private static async Task<List<Activity>> ReadBotMessagesAsync(DirectLineClient client, string conversationId)
+        {
+            string watermark = null;
```
peterinnesmsft: Watermark isn't being used - remove.

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            List<Activity> botResponses = null;
+
+            // Poll the bot for replies once per second.
```
peterinnesmsft: Remove this comment - polling occurs outside of this function so it's confusing for this this be located here.

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
```diff
@@ -0,0 +1,231 @@
...
+                    await Task.Delay(TimeSpan.FromSeconds(20)).ConfigureAwait(false);
+                }
+            }
+            while (responses?.FirstOrDefault()?.Text == null || allFirstPromptTextVariations.Contains(responses?.FirstOrDefault()?.Text));
+
+            return responses;
+        }
+
+        /// <summary>
+        /// Polls the bot continuously until it gets a response.
```
peterinnesmsft: Again, this function is really just retrieving activities for the conversation. It's not polling itself.

templates/typescript/samples/sample-skill/functional-tests/SkillSample.FunctionalTests/DirectLineClientTests.cs
peterinnesmsft: Although, is there anything that would cause the bot ID to not match? Do we need to actually need to do this check?

peterinnesmsft (changes requested):
Let's take a look at the test implementation and tweak it a little. I'd also like to better understand the periodic errors that are initially returned by a TypeScript bot, and fix that correctly rather than work around the issues in the test.

Batta32:
Thanks @peterinnesmsft, we will review the entire test implementation.

We converted this PR to draft and as soon as it's ready we will move it to be reviewed again.

Batta32:
@peterinnesmsft - we already applied the requested feedback.

The next week we will validate the changes in order to confirm that the Functional Test with the TypeScript Skill are working as expected.

Feel free to add the necessary feedback!

Batta32:
Hi @peterinnesmsft, we successfully **validated** the changes running the functional test in a pipeline created by the yml. The PR is ready to add any further review or merge.

We will be applying the same feedback for the PR #3463 and PR #3602.

image

lauren-mills:
Closing - outdated

PR #3559, opened 2020-07-15
[TypeScript][Virtual Assistant & Skill] Implement Dependency Injection

### LINKED ISSUE #3496: QnAMakerDialog is not created for different languages

Virtual Assistant and Skills

Typescript 

#### What happens?

When a **QnAMakerDialog** is created with a specific language (ex. DE) will be resolving always that language even though you change the language or another user uses a different language (ex EN).

**User 1**: Sends utterance to the bot ( DE );
**Bot**: The Bot resolves the utterance as QnA intent, and create the new QnADialog, with the knowledgebaseId as Id of the Dialog:

   ```
 private registerQnADialog(knowledgebaseId: string, cognitiveModels: ICognitiveModelSet, locale: string): void {
        const qnaEndpoint: QnAMakerEndpoint | undefined = cognitiveModels.qnaConfiguration.get(knowledgebaseId);
        if (qnaEndpoint == undefined){
            throw new Error(`Could not find QnA Maker knowledge base configuration with id: ${ knowledgebaseId }.`);
        }

        if (this.dialogs.find(knowledgebaseId) == undefined) {
            const qnaDialog: QnAMakerDialog = new QnAMakerDialog(
                qnaEndpoint.knowledgeBaseId,
                qnaEndpoint.endpointKey,
                // The following line is a workaround until the method getQnAClient of QnAMakerDialog is fixed
                // as per issue https://github.com/microsoft/botbuilder-js/issues/1885
                new URL(qnaEndpoint.host).hostname.split('.')[0],
                this.templateEngine.generateActivityForLocale('UnsupportedMessage', locale) as Activity,
                // Before, instead of 'undefined' a '0.3' value was used in the following line
                undefined,
                this.templateEngine.generateActivityForLocale('QnaMakerAdaptiveLearningCardTitle', locale).text,
                this.templateEngine.generateActivityForLocale('QnaMakerNoMatchText', locale).text
            );

            qnaDialog.id = knowledgebaseId;

            this.addDialog(qnaDialog);
        }
    }
``` 

**Bot**: Sends QnA Answer to User 1 in User 1's locale ( DE )
**User 2**: Sends utterance to bot ( EN );
**Bot**: The Bot resolves the utterance as QnA intent ( same the user 1 sent ), no new QnaMakerDialog is created because one Dialog with the same knowledgebaseId already exists.
**Bot**: The Bot start again the QnA Dialog (added from user 1 iteration) and get the answer from the Knowledge Base (DE) of the first user;
**Bot**: Sends Message to User 2 in User 1's locale.

Another **QnAMakerDialog** will be created with the right language and right knowledge base ID.

### PR DESCRIPTION
**Pull request by Batta32**
Fix #3496 

### Purpose
We introduced `inversifyjs` library in order to add Dependency Injection functionality following parity with C#.

### Changes
The following changes were replicated in Virtual Assistant and Skill:

- Refactor of index/dialogs folder in order to integrate Dependency Injection
- Update `tsconfig.json` to be able to use Dependency Injection
- Add `constant` and `inversify.config.ts` files to configure Dependency Injection functionality
- Update `package.json` adding `inversifyjs` and `reflect-metadata` libraries
- Fix tests VA/Skill and generator
- Replicate changes in template

### Tests
\-

### Feature Plan
\-

#### General
- [X] I have commented my code, particularly in hard-to-understand areas	

### DIFF
--- a/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/adapters/defaultAdapter.ts
+++ b/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/adapters/defaultAdapter.ts
@@ -23,9 +23,10 @@ import {
     FeedbackOptions} from 'bot-solutions';
 import { TelemetryInitializerMiddleware } from 'botbuilder-applicationinsights';
 import { IBotSettings } from '../services/botSettings';
-import { inject } from 'inversify';
+import { inject, injectable } from 'inversify';
 import { TYPES } from '../types/constants';
 
+@injectable()
 export class DefaultAdapter extends BotFrameworkAdapter {
 
     public constructor(@inject(TYPES.BotSettings) settings: Partial<IBotSettings>,
--- a/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/adapters/defaultAdapter.ts
+++ b/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/adapters/defaultAdapter.ts
@@ -28,8 +28,7 @@ import { TYPES } from '../types/constants';
 
 export class DefaultAdapter extends BotFrameworkAdapter {
 
-    public constructor(
-    @inject(TYPES.BotSettings) settings: Partial<IBotSettings>,
+    public constructor(@inject(TYPES.BotSettings) settings: Partial<IBotSettings>,
         @inject(TYPES.LocaleTemplateManager) templateManager: LocaleTemplateManager,
         @inject(TYPES.ConversationState) conversationState: ConversationState,
         @inject(TYPES.BotFrameworkAdapterSettings) adapterSettings: Partial<BotFrameworkAdapterSettings>,
--- a/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/adapters/defaultAdapter.ts
+++ b/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/adapters/defaultAdapter.ts
@@ -22,17 +22,19 @@ import {
     SetSpeakMiddleware,
     FeedbackOptions} from 'bot-solutions';
 import { TelemetryInitializerMiddleware } from 'botbuilder-applicationinsights';
-import { IBotSettings } from '../services/botSettings.js';
+import { IBotSettings } from '../services/botSettings';
+import { inject } from 'inversify';
+import { TYPES } from '../types/constants';
 
 export class DefaultAdapter extends BotFrameworkAdapter {
 
     public constructor(
-        settings: Partial<IBotSettings>,
-        templateManager: LocaleTemplateManager,
-        conversationState: ConversationState,
-        adapterSettings: Partial<BotFrameworkAdapterSettings>,
-        telemetryMiddleware: TelemetryInitializerMiddleware,
-        telemetryClient: BotTelemetryClient
+    @inject(TYPES.BotSettings) settings: Partial<IBotSettings>,
+        @inject(TYPES.LocaleTemplateManager) templateManager: LocaleTemplateManager,
+        @inject(TYPES.ConversationState) conversationState: ConversationState,
+        @inject(TYPES.BotFrameworkAdapterSettings) adapterSettings: Partial<BotFrameworkAdapterSettings>,
+        @inject(TYPES.TelemetryInitializerMiddleware) telemetryMiddleware: TelemetryInitializerMiddleware,
+        @inject(TYPES.BotTelemetryClient) telemetryClient: BotTelemetryClient
     ) {
         super(adapterSettings);
 
--- a/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/adapters/defaultAdapter.ts
+++ b/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/adapters/defaultAdapter.ts
@@ -43,10 +43,12 @@ export class DefaultAdapter extends BotFrameworkAdapter {
                 type: ActivityTypes.Trace,
                 text: error.message
             });
+
             await context.sendActivity({
                 type: ActivityTypes.Trace,
                 text: error.stack
             });
+
             await context.sendActivity(templateManager.generateActivityForLocale('ErrorMessage', context.activity.locale));
             telemetryClient.trackException({ exception: error });
         };
--- a/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
+++ b/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
@@ -59,16 +59,13 @@ container.bind<BotTelemetryClient>(TYPES.BotTelemetryClient).toConstantValue(
 );
 
 decorate(injectable(), TelemetryLoggerMiddleware);
-container.bind<TelemetryLoggerMiddleware>(TYPES.TelemetryLoggerMiddleware).toConstantValue(
-    new TelemetryLoggerMiddleware(container.get<BotTelemetryClient>(TYPES.BotTelemetryClient))
-);
+decorate(inject(TYPES.BotTelemetryClient), TelemetryLoggerMiddleware, 0);
+container.bind<TelemetryLoggerMiddleware>(TYPES.TelemetryLoggerMiddleware).to(TelemetryLoggerMiddleware).inSingletonScope();
 
 decorate(injectable(), TelemetryInitializerMiddleware);
-container.bind<TelemetryInitializerMiddleware>(TYPES.TelemetryInitializerMiddleware).toConstantValue(
-    new TelemetryInitializerMiddleware(
-        container.get<TelemetryLoggerMiddleware>(TYPES.TelemetryLoggerMiddleware)
-    )
-);
+decorate(inject(TYPES.TelemetryLoggerMiddleware), TelemetryInitializerMiddleware, 0);
+container.bind<TelemetryInitializerMiddleware>(TYPES.TelemetryInitializerMiddleware).to(TelemetryInitializerMiddleware).inSingletonScope();
+
 
 // Configure bot services
 container.bind<BotServices>(TYPES.BotServices).to(BotServices).inSingletonScope();
@@ -79,22 +76,16 @@ container.bind<BotServices>(TYPES.BotServices).to(BotServices).inSingletonScope(
 // container.bind<Partial<MemoryStorage>>(TYPES.MemoryStorage).toConstantValue(new MemoryStorage());
 decorate(injectable(), CosmosDbPartitionedStorage);
 container.bind<CosmosDbPartitionedStorage>(TYPES.CosmosDbPartitionedStorage).toConstantValue(
-    new CosmosDbPartitionedStorage(
-        container.get<IBotSettings>(TYPES.BotSettings).cosmosDb
-    )
+    new CosmosDbPartitionedStorage(container.get<IBotSettings>(TYPES.BotSettings).cosmosDb)
 );
 
 decorate(injectable(), UserState);
-container.bind<UserState>(TYPES.UserState).toConstantValue(
-    new UserState(
-        container.get<CosmosDbPartitionedStorage>(TYPES.CosmosDbPartitionedStorage)
-    )
-);
+decorate(inject(TYPES.CosmosDbPartitionedStorage), UserState, 0);
+container.bind<UserState>(TYPES.UserState).to(UserState).inSingletonScope();
 
 decorate(injectable(), ConversationState);
-container.bind<ConversationState>(TYPES.ConversationState).toConstantValue(
-    new ConversationState(container.get<CosmosDbPartitionedStorage>(TYPES.CosmosDbPartitionedStorage))
-);
+decorate(inject(TYPES.CosmosDbPartitionedStorage), ConversationState, 0);
+container.bind<ConversationState>(TYPES.ConversationState).to(ConversationState).inSingletonScope();
 
 // Configure localized responses
 const supportedLocales: string[] = ['en-us', 'de-de', 'es-es', 'fr-fr', 'it-it', 'zh-cn'];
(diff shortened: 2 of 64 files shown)

### CODE REVIEW

**Review by lzc850612 — APPROVED** [review]

File: templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/adapters/defaultAdapter.ts [hunk 1]
```diff
@@ -22,17 +22,19 @@ import {
...
 export class DefaultAdapter extends BotFrameworkAdapter {
 
     public constructor(
-        settings: Partial<IBotSettings>,
-        templateManager: LocaleTemplateManager,
-        conversationState: ConversationState,
-        adapterSettings: Partial<BotFrameworkAdapterSettings>,
-        telemetryMiddleware: TelemetryInitializerMiddleware,
-        telemetryClient: BotTelemetryClient
+    @inject(TYPES.BotSettings) settings: Partial<IBotSettings>,
```
peterinnesmsft: Fix indentation.

peterinnesmsft: Seeing quite a lot of places where this is true in the constructor - please fix throughout where applicable.

templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
```diff
@@ -0,0 +1,246 @@
...
+);
+decorate(injectable(), TelemetryInitializerMiddleware);
+container.bind<TelemetryInitializerMiddleware>(TYPES.TelemetryInitializerMiddleware).toConstantValue(
+    new TelemetryInitializerMiddleware(
+        container.get<TelemetryLoggerMiddleware>(TYPES.TelemetryLoggerMiddleware)
+    )
+);
+
+// Configure bot services
+decorate(injectable(), BotServices);
```
peterinnesmsft: BotServices is defined within our scope - why not just add `@injectable` at the class definition? This method seems appropriate for objects we're pulling from external packages, but for things we control, I would assume we could just add the decorator in the class directly.

peterinnesmsft: Same for any other classes defined within our area.

templates/typescript/generator-bot-virtualassistant/generators/skill/templates/sample-skill/src/dialogs/skillDialogBase.ts
```diff
@@ -31,12 +33,12 @@ export class SkillDialogBase extends ComponentDialog {
     protected templateManager: LocaleTemplateManager;
 
     public constructor(
-        dialogId: string,
-        settings: Partial<IBotSettings>,
-        services: BotServices,
-        stateAccessor: StatePropertyAccessor<SkillState>,
-        telemetryClient: BotTelemetryClient,
-        templateManager: LocaleTemplateManager
+    @unmanaged() dialogId: string,
```
peterinnesmsft: Do we actually need this here? My expectation is that we should never actually be instantiating SkillDialogBase, but rather some derived class of it. It makes sense to use injection in the child, but I would then assume it would just pass injected parameters through to the parent constructor using super, which would mean these ultimately never get injected?

templates/typescript/samples/sample-skill/src/bots/defaultActivityHandler.ts
```diff
@@ -2,7 +2,7 @@
  * Copyright(c) Microsoft Corporation.All rights reserved.
  * Licensed under the MIT License.
  */
-
+import 'reflect-metadata';
```
peterinnesmsft, line 6: new line after copyright header

templates/typescript/samples/sample-skill/src/dialogs/skillDialogBase.ts
```diff
@@ -31,12 +33,12 @@ export class SkillDialogBase extends ComponentDialog {
     protected templateManager: LocaleTemplateManager;
 
     public constructor(
-        dialogId: string,
-        settings: Partial<IBotSettings>,
-        services: BotServices,
-        stateAccessor: StatePropertyAccessor<SkillState>,
-        telemetryClient: BotTelemetryClient,
-        templateManager: LocaleTemplateManager
+    @unmanaged() dialogId: string,
```
peterinnesmsft: Same thing here - do we need to add this on the parent if we should never be instantiating this directly?

templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
```diff
@@ -0,0 +1,246 @@
...
+        container.get<SkillConversationIdFactoryBase>(TYPES.SkillConversationIdFactory)
+    )
+);
+
+// Register dialogs
+decorate(injectable(), Dialog);
+decorate(injectable(), DialogContainer);
+decorate(injectable(), ComponentDialog);
+decorate(injectable(), MainDialog);
+container.bind<MainDialog>(TYPES.MainDialog).to(MainDialog).inTransientScope();
```
peterinnesmsft, line 130: Do we need .inTransientScope given that it is the default behavior? Seems inconsistent to have it here but not for everything else that is transient. From a readability perspective to someone who is new and not so familiar with InversifyJS, it may make sense to include it.

templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
```diff
@@ -0,0 +1,246 @@
...
+    blobStorage: appsettings.blobStorage,
+    cognitiveModels: cognitiveModels,
+    cosmosDb: appsettings.cosmosDb,
+    defaultLocale: cognitiveModelsRaw.defaultLocale,
+    microsoftAppId: appsettings.microsoftAppId,
+    microsoftAppPassword: appsettings.microsoftAppPassword
+};
+
+// Load settings
+container.bind<Partial<IBotSettings>>(TYPES.BotSettings).toConstantValue(botSettings);
```
peterinnesmsft, line 48: How does scope behavior work with constants? As in, should we optimize this to use singleton scope if it's not changing? I've not used the library before so I'm not sure of the behavior, but would be good to check this.

templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
```diff
@@ -0,0 +1,246 @@
...
+// Configure configuration provider
+decorate(injectable(), SimpleCredentialProvider);
+container.bind<SimpleCredentialProvider>(TYPES.SimpleCredentialProvider).toConstantValue(
+    new SimpleCredentialProvider(appsettings.microsoftAppId, appsettings.microsoftAppPassword)
+);
+
+// Configure telemetry
+container.bind<BotTelemetryClient>(TYPES.BotTelemetryClient).toConstantValue(
+    getTelemetryClient(container.get<IBotSettings>(TYPES.BotSettings))
+);
```
peterinnesmsft, line 59: Being kind of inconsistent with styling here - general rule of thumb, new lines after closing ); if it's on its own line.

templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
```diff
@@ -0,0 +1,246 @@
...
+decorate(injectable(), AuthenticationConfiguration);
+
+// Register the SkillDialogs (remote skills).
+const skills: IEnhancedBotFrameworkSkill[] = appsettings.botFrameworkSkills;
+if (skills !== undefined && skills.length > 0) {
+    const hostEndpoint: string = appsettings.skillHostEndpoint;
+    if (hostEndpoint === undefined || hostEndpoint.trim().length === 0) {
+        throw new Error('\'skillHostEndpoint\' is not in the configuration');
+    } else {
+        decorate(injectable(), SkillsConfiguration);
```
peterinnesmsft: Adding a decorator conditionally seems weird to me. I get it...just feels funny.

templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
```diff
@@ -0,0 +1,246 @@
...
+
+decorate(injectable(), AuthenticationConfiguration);
+
+// Register the SkillDialogs (remote skills).
+const skills: IEnhancedBotFrameworkSkill[] = appsettings.botFrameworkSkills;
+if (skills !== undefined && skills.length > 0) {
+    const hostEndpoint: string = appsettings.skillHostEndpoint;
+    if (hostEndpoint === undefined || hostEndpoint.trim().length === 0) {
+        throw new Error('\'skillHostEndpoint\' is not in the configuration');
+    } else {
```
peterinnesmsft: No need for else since you're throwing above.

templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
```diff
@@ -0,0 +1,246 @@
...
+decorate(injectable(), TeamsActivityHandler);
+decorate(injectable(), DefaultActivityHandler);
+container.bind<DefaultActivityHandler<MainDialog>>(TYPES.DefaultActivityHandler).to(DefaultActivityHandler).inTransientScope();
+
+// Configure SkillHandler
+decorate(injectable(), ChannelServiceHandler);
+decorate(injectable(), SkillHandler);
+
+container.bind<SkillHandler>(TYPES.SkillHandler).toConstantValue(new SkillHandler(
+    container.get<DefaultAdapter>(TYPES.DefaultAdapter),
```
peterinnesmsft: This seems weird...isn't the whole point that you shouldn't have to do this? As long as you have marked the classes needed for the constructor here as `@injectable`, then when you access SkillHandler from the container, it should be able to automatically get the necessary dependencies, no?

peterinnesmsft (changes requested):
Some formatting to fixes to apply, and also just some general questions about InversifyJS since I'm not overly familiar with it.

Batta32:
Thanks @peterinnesmsft for the feedback! We will be reviewing this as soon as possible.

templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
Batta32: @peterinnesmsft - as you mentioned, this is the [default behavior](https://github.com/inversify/InversifyJS/blob/master/wiki/scope.md#controlling-the-scope-of-the-dependencies), however, this was included to improve the readability of the scopes for each class and to follow the parity with [C#](https://github.com/microsoft/botframework-solutions/blob/master/samples/csharp/assistants/virtual-assistant/VirtualAssistantSample/Startup.cs#L124).

Let us know if we should the remove `.inTransientScope()` of those classes.

templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
Batta32: @peterinnesmsft - we researched about the `toConstantValue()` vs `inSingletonScope()`.
* `toConstantValue()`: injects a value (it will not create an instance) as a singleton.
* `inSingletonScope()`: will create an instance and inject it as a singleton.

We double-checked the implementation of `toConstantValue()` in this PR and noticed that it is implemented for _interfaces_ and those classes that are _outside of our scope_ (e.g `SkillDialog`).

Changing from `toConstantValue()` to `inSingletonScope()` requires some research and implementation for the classes that are outside of our scope.

Let us know if we should create a plan to do the migration of `toConstantValue()` to `inSingletonScope()`.

Documentation:
* [Bindings that will inject an object](https://github.com/inversify/InversifyJS/blob/master/wiki/scope.md#bindings-that-will-inject-an-object)
* [Question](https://stackoverflow.com/questions/41105753/how-to-create-a-injected-and-a-const-instance/41105948#41105948) in stackoverflow

Batta32:
@peterinnesmsft - we already applied the requested changes mentioned above!

templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/inversify.config.ts
Batta32: We replaced most of the `toConstantValue()` to `inSingletonScope()` in the Virtual Assistant and in the Skill. We will plan to replace the remaining ones in the next week!

templates/typescript/generator-bot-virtualassistant/generators/skill/templates/sample-skill/src/bots/defaultActivityHandler.ts
```diff
@@ -15,18 +16,18 @@ import {
...
     private readonly dialog: Dialog;
     private readonly conversationState: BotState;
     private readonly userState: BotState;
     private dialogStateAccessor: StatePropertyAccessor<DialogState>;
     private templateManager: LocaleTemplateManager;
 
-    public constructor(
-    @inject(TYPES.ConversationState) conversationState: BotState,
+    public constructor(@inject(TYPES.ConversationState) conversationState: BotState,
         @inject(TYPES.ConversationState) userState: BotState,
```
tomSauret847: Line 31 has a typo. It is saving Conversation State to the User State

Batta32: Thanks @tomSauret847! We already fixed it.

tomSauret847: Thank you @Batta32 I am building another VA and set of skills using this branch for the base.

PR #3628, opened 2020-08-19
Update DI scope for DefaultAdapter from Transient to Singleton

### LINKED ISSUE #3627: BotFrameworkAdapter should be a singleton
SkillSample

C#

#### What happens?
We're incorrectly setting up the BotFrameworkAdapter, which is causing auth token caching to be disabled.

Refer to:
https://github.com/microsoft/botframework-solutions/blob/970aaf144f06b6695bbf77ad5707344f825b92e6/samples/csharp/skill/SkillSample/Startup.cs#L124

peterinnesmsft:
### Purpose
Fixes #3627.

The underlying implementation for DefaultAdapter (BotFrameworkAdapter) contains a credential cache which is used to refresh credentials approximately every 30 minutes. However, DefaultAdapter is added to DI services using the Transient scope, which means that the adapter is instantiated on each request, which forces acquisition of the credentials and defeats the point of the cache. This change modifies the scope of DefaultAdapter to Singleton so it is appropriately reused across all requests.

### Changes
- Update templates/csharp/Skill/Skill/Startup.cs to change DI scope of DefaultAdapter from Transient to Singleton.
- Update samples/csharp/skill/SkillSample/Startup.cs to change DI scope of DefaultAdapter from Transient to Singleton.

### Tests
N/A

### Feature Plan
N/A

#### General

### DIFF
--- a/samples/csharp/skill/SkillSample/Startup.cs
+++ b/samples/csharp/skill/SkillSample/Startup.cs
@@ -131,7 +131,7 @@ namespace SkillSample
             services.AddTransient<MainDialog>();
 
             // Configure adapters
-            services.AddTransient<IBotFrameworkHttpAdapter, DefaultAdapter>();
+            services.AddSingleton<IBotFrameworkHttpAdapter, DefaultAdapter>();
 
             // Configure bot
             services.AddTransient<IBot, DefaultActivityHandler<MainDialog>>();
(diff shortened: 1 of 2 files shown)


lauren-mills (approved):

ryanisgrig (approved):

taicchoumsft (approved):
:shipit:

PR #3633, opened 2020-08-19
remove checking on activity.from.role

### LINKED ISSUE #3610: Proactive State document not updated
BotFramework Solutions VA Template(v1.0.1), BotBuilder 4.9.3
C#
#### What happens?
The Proactive state document in Cosmos DB does not get immediately updated with the latest conversation details. Due to this, any proactive notifications are not sent to the right conversation. I face this issue when I connect through WebChat, bot in my application, as well as when I access the bot through the "Test in WebChat option" on Azure. Instead, the details are only updated once I invoke a skill dialog. On the Bot Framework Emulator however , there is an immediate update of the conversation details in the ProactiveState document. 

In the new VA template, the **CosmosDbStorage** class has been replaced by **CosmosDbPartitionedStorage** class, and BotSettings now takes **CosmosDbPartitionedStorageOptions** as opposed to **CosmosDbStorageOptions** in older versions. Could this be causing an issue?

1. Set up Proactive Notifications handling from [here](https://microsoft.github.io/botframework-solutions/solution-accelerators/tutorials/enable-proactive-notifications/4-virtual-assistant/)
2. Start a new conversation with the VA.
3. Query the ProactiveState document in CosmosDB.
4. The conversation ID entry for the given userID has an older value as compared to the current conversationId when tested from WebChat, but has updated value if tested from Emulator.

The Proactive state document must be updated as soon as a new conversation is started, to the latest conversation details.

### LINKED ISSUE #3614: "Role" property of conversation update activity set to null
## Version
WebChat 4.9.1 via CDN

## Describe the bug

When the conversationUpdate activity is sent by WebChat, the Role property of the activity is set to null. As a result, the ProactiveState does not get updated with new Conversation details since it looks for the role being equal to "user". This gets updated only when a skill is invoked/message is sent. As a result, if there are any proactive notifications, they are sent to the wrong conversation.

## Steps to reproduce

1. Initialize connection to Bot through WebChat using DirectLine.
2. Open a connection to the Bot.
3. Note in the bot back-end logic that turnContext.Activity.From.Role is set to null, and not "user".
4. Note the error.

## Expected behavior

WebChat should pass the relevant role as "user", so that the Proactive State functionalities are not broken.

## Additional context

WebChat implemented in Angular application.

[Bug]

lzc850612:
### Purpose

For ProactiveMiddleware, we do not need to check if an activity is from user per the latest botbuilder SDK.

### Changes

Remove checking on activity.from.role and only check activity.from.id

### Tests

### Feature Plan

#### General

### DIFF
--- a/sdk/csharp/libraries/microsoft.bot.solutions/Proactive/ProactiveStateMiddleware.cs
+++ b/sdk/csharp/libraries/microsoft.bot.solutions/Proactive/ProactiveStateMiddleware.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Bot.Solutions.Proactive
         {
             var activity = turnContext.Activity;
 
-            if (!string.IsNullOrEmpty(activity.From.Role) && activity.From.Role.Equals("user", StringComparison.InvariantCultureIgnoreCase))
+            if (!string.IsNullOrWhiteSpace(activity.From?.Id))
             {
                 var proactiveState = await _proactiveStateAccessor.GetAsync(turnContext, () => new ProactiveModel()).ConfigureAwait(false);
                 ProactiveModel.ProactiveData data;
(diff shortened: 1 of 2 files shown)


EricDahlvang (approved):
lgtm

peterinnesmsft (approved):

PR #3637, opened 2020-08-24
[PORT #3262][Botskills] Show all language output when there's an issue with a LU file

Batta32:
Port #3262 

### Purpose

At the moment, when botskills processes multiple languages and an error occurs while parsing LU files it will stop processing any subsequent LU file, therefore no further error will be logged.

With these changes, when there is an issue while parsing multiple languages with Ludown it will show all the error messages if there are issues with more than one, instead of showing only the first error message.

image

### Changes

The exception throwing in `executeLudownParse` and `executeDispatchAdd` is changed to error loging.
After processing each language, if there was an error it will throw an exception.

### Tests
image

### Feature Plan

#### General

### DIFF
--- a/tools/botskills/src/functionality/connectSkill.ts
+++ b/tools/botskills/src/functionality/connectSkill.ts
@@ -278,7 +278,7 @@ Make sure you have a Dispatch for the cultures you are trying to connect, and th
                 }));
 
             if (this.logger.isError) {
-                throw new Error(`There were issues while converting the LU files.`);
+                throw new Error('There were one or more issues converting the LU files. Aborting the process.');
             }
 
             // Check if it is necessary to refresh the skill
--- a/tools/botskills/src/functionality/connectSkill.ts
+++ b/tools/botskills/src/functionality/connectSkill.ts
@@ -214,7 +214,7 @@ Make sure you have a Dispatch for the cultures you are trying to connect, and th
                 throw new Error(`Path to ${ luisFile } (${ luisFilePath }) leads to a nonexistent file.`);
             }
         } catch (err) {
-            throw new Error(`There was an error in the bf luis:convert command:\nCommand: ${ luisConvertCommand.join(' ') }\n${ err }`);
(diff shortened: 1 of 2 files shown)


tools/botskills/src/functionality/connectSkill.ts
```diff
@@ -272,9 +272,15 @@ Make sure you have a Dispatch for the cultures you are trying to connect, and th
...
                     const executionModelByCulture: Map<string, string> = item[1];
                     await this.executeLuisConvert(culture, executionModelByCulture);
-                    await this.executeDispatchAdd(culture, executionModelByCulture);
+                    if (!this.logger.isError) {
+                        await this.executeDispatchAdd(culture, executionModelByCulture);
+                    }
                 }));
 
+            if (this.logger.isError) {
+                throw new Error(`There were issues while converting the LU files.`);
```
peterinnesmsft: Where does logger output the errors to - is it the console when running from cmd for example? If not, we may want to add some kind of message that generally indicates check log output for details, otherwise this will probably end up confusing and frustrating users if they can't easily determine what went wrong.

peterinnesmsft (approved):
Consider updating the thrown error message if any errors occur, but otherwise looks good to me. :shipit:

Batta32:
Thanks @peterinnesmsft - we will be reviewing this and we will back here as soon as we can.

taicchoumsft (approved):
:shipit:

Batta32:
@peterinnesmsft - at the moment Botskills is showing **all the messages in the console** during the execution, as it does not log on files.

We analyzed possible improvements and determined they would require further changes in Botskills. So we came up with an **action plan**:
1. **Move** this Pull Request to draft
1. **Improve** the structure of the error messages

The improvements would entail removing try catch blocks were possible. We noticed that this is causing a stack of error messages as exceptions are thrown in low level methods and start being caught to the top.

As a **default action**, the next week we will start implementing the action plan detailed above.

peterinnesmsft (approved):

PR #3740, opened 2020-12-13
display correct message with current locale

chenxizhang:
### Purpose
I found this issue with my project. I created two languages models, and I found if any intent was matched, there will always show english fallback message, not my current locale (for example zh-cn),
### Changes
 I checked the code, and added few changes, then it works fine for me.

### Tests

### Feature Plan

#### General

### DIFF
--- a/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/dialogs/mainDialog.ts
+++ b/templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/dialogs/mainDialog.ts
@@ -399,7 +399,7 @@ export class MainDialog extends ComponentDialog {
     private async finalStep(stepContext: WaterfallStepContext): Promise<DialogTurnResult> {
         // Clear active skill in state.
         await this.activeSkillProperty.delete(stepContext.context);
-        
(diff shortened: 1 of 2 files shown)


templates/typescript/generator-bot-virtualassistant/generators/app/templates/sample-assistant/src/dialogs/mainDialog.ts
```diff
@@ -399,7 +399,7 @@ export class MainDialog extends ComponentDialog {
     private async finalStep(stepContext: WaterfallStepContext): Promise<DialogTurnResult> {
         // Clear active skill in state.
         await this.activeSkillProperty.delete(stepContext.context);
-        
+
```
chenxizhang, line 402: This line is not necessery

Batta32:
Hi @chenxizhang! Sorry for the delay. We will close this PR as we noticed that these were applied in the PR #3411 in `next` branch and they will be available in `master` soon.

You can see those changes applied [here](https://github.com/microsoft/botframework-solutions/blob/next/templates/typescript/samples/sample-assistant/src/dialogs/mainDialog.ts#L214-L217).

